PR #67, opened 2015-06-25
Enable packaging with https://packager.io

crohr:
I saw you tried packaging on https://packager.io, but were missing the [`.godir` file](https://packager.io/documentation/golang/). So here is a working `.pkgr.yml` file for some of the supported distributions.

Packages can be found at https://packager.io/gh/pkgr/scaleway-cli/install?bid=2#ubuntu-14-04-scw, and I verified that you can install and run the `scw` command on Ubuntu 14.04, so it should work on the other distributions as well.

Let me know if this is of interest to you! We can also provide a standalone server so that you can serve the generated packages from your own domain and servers.

### DIFF
--- a/.pkgr.yml
+++ b/.pkgr.yml
@@ -1,0 +1,11 @@
+cli: scw
+targets:
+  ubuntu-14.04:
+  centos-7:
+  ubuntu-12.04:
+  centos-6:
+  debian-7:
+before:
+  - echo "scw" > .godir
+after:
+  - rm -f .godir


moul:
Nice :)

moul:
Do you have or plan to have Github Badge like  ?

crohr:
@moul that would be nice indeed, but haven't had the time to implement it yet. I'll give it a go as soon as I can ;)

moul:
Ok, we will do our own for now :)

Also, I suggest you to add the ubuntu 15.04 distribution as it's the first official ubuntu with systemd by default

crohr:
At this point we're only supporting LTS releases for Ubuntu. Support for systemd is already working with Debian 8 and RHEL/CentOS 7 so it won't be an issue once the new LTS comes.

pataquets:
As of now, this seems broken.
@moul : Filed issue #1534.

PR #76, opened 2015-07-21
Rename scaleway-cli into scw at the end of packaging

crohr:
Now that the `.godir` uses `scaleway-cli` as the binary name, we need to rename it at the end of the compilation so that one can successfully run `scw` after the package is installed.


moul:
Perfect, thanks

PR #106, opened 2015-08-07
add account api in the config file

QuentinPerez:


cmd/scw/main.go
```diff
@@ -157,6 +157,11 @@ func getConfig() (*api.Config, error) {
 	if err != nil {
 		return nil, err
 	}
+	// check if he has an old scwrc version
+	if config.ACCOUNTEndPoint == "" {
```
moul: `s/ACCOUNTEndPoint/AccountEndpoint/g`

pkg/api/api.go
```diff
@@ -500,17 +506,19 @@ var FuncMap = template.FuncMap{
 }
 
 // NewScalewayAPI creates a ready-to-use ScalewayAPI client
-func NewScalewayAPI(endpoint, organization, token string) (*ScalewayAPI, error) {
+func NewScalewayAPI(apiEndPoint, accountEndPoint, organization, token string) (*ScalewayAPI, error) {
```
moul: Can you run a `make test`, you will need to update the test in ./pkg/commands/command_test.go

pkg/api/api.go
```diff
@@ -500,17 +506,19 @@ var FuncMap = template.FuncMap{
...
 	}
 	s := &ScalewayAPI{
-		APIEndPoint:  endpoint,
-		Organization: organization,
-		Token:        token,
-		Cache:        cache,
-		anonuuid:     *anonuuid.New(),
+		APIEndPoint:     apiEndPoint,
+		ACCOUNTEndPoint: accountEndPoint,
+		APIurl:          apiEndPoint,
```
moul: I suggest you to remove the `APIurl` hack with the `enable`/`disable` methods, we will hardcode which API to use depending on the resource we want to work with

pkg/utils/utils.go
```diff
@@ -168,15 +169,11 @@ func RemoveDuplicates(elements []string) []string {
 
 // GetConfigFilePath returns the path to the Scaleway CLI config file
 func GetConfigFilePath() (string, error) {
-	homeDir := os.Getenv("HOME") // *nix
-	if homeDir == "" {           // Windows
-		homeDir = os.Getenv("USERPROFILE")
-	}
-	if homeDir == "" {
-		return "", errors.New("user home directory not found")
+	usr, err := user.Current()
```
moul: `user.Current()` forbid the creation of static binaries when cross compiling, so we need to use the oldschool way :)

pkg/commands/login.go
```diff
@@ -34,36 +32,22 @@ func RunLogin(ctx CommandContext, args LoginArgs) error {
...
+		Organization:    strings.Trim(args.Organization, "\n"),
+		Token:           strings.Trim(args.Token, "\n"),
 	}
 
-	api, err := api.NewScalewayAPI(cfg.APIEndPoint, cfg.Organization, cfg.Token)
+	api, err := api.NewScalewayAPI(cfg.APIEndPoint, cfg.ACCOUNTEndPoint, cfg.Organization, cfg.Token)
 	if err != nil {
 		return fmt.Errorf("Unable to create ScalewayAPI: %s", err)
 	}
+	api.EnableAccountApi()
```
moul: Can you move this call in `api.CheckCredentials()`, and calls `DisableAccountApi` just after but still inside the same function ?

pkg/api/api.go
moul: Ok, got it, you can keep this hack to keep only one instance of each functions :)

pkg/api/api.go
```diff
@@ -1193,6 +1201,8 @@ func (s *ScalewayAPI) GetTasks() (*[]ScalewayTask, error) {
 
 // CheckCredentials performs a dummy check to ensure we can contact the API
 func (s *ScalewayAPI) CheckCredentials() error {
+	s.enableAccountApi()
+	defer s.disableAccountApi()
```
moul: :+1:

pkg/api/config.go
```diff
@@ -0,0 +1,46 @@
...
+		return fmt.Errorf("Unable to create scwrc config file: %s", err)
+	}
+	defer scwrc.Close()
+	encoder := json.NewEncoder(scwrc)
+	err = encoder.Encode(c)
+	if err != nil {
+		return fmt.Errorf("Unable to encode scw config file: %s", err)
+	}
+	return nil
+}
```
moul: :+1:

moul:
LGTM

PR #107, opened 2015-08-08
upload ssh key

QuentinPerez:


pkg/api/api.go
```diff
@@ -1210,11 +1269,37 @@ func (s *ScalewayAPI) CheckCredentials() error {
...
 	}
 	if resp.StatusCode != 200 {
-		return fmt.Errorf("invalid credentials")
+		return fmt.Errorf("[%d] invalid credentials", resp.StatusCode)
 	}
 	return nil
 }
 
+// GetUUID returns the UUID
+func (s *ScalewayAPI) GetUUID() (string, error) {
```
moul: You should rename `GetUUID` and `UUID` to `GetUserID` and `UserID`, UUID is too much *universal* for this context

pkg/commands/login.go
```diff
@@ -7,23 +7,84 @@ package commands
...
+		args.IndexSSH = string(buff[:len(buff)])
+		break
+	}
+	return nil
 }
 
 // RunLogin is the handler for 'scw login'
 func RunLogin(ctx CommandContext, args LoginArgs) error {
+	if args.IndexSSH == "" {
+		if err := selectKey(&args); err != nil {
```
moul: You should ask for user SSH key after calling `apiConnection.checkCredentials()`

pkg/commands/login.go
```diff
@@ -7,23 +7,84 @@ package commands
...
 
 	"github.com/scaleway/scaleway-cli/pkg/api"
+	"github.com/scaleway/scaleway-cli/pkg/utils"
 )
 
 // LoginArgs are arguments passed to `RunLogin`
 type LoginArgs struct {
 	Organization string
 	Token        string
+	IndexSSH     string
```
moul: Why are you using `Index` in this variable name ?

pkg/commands/login.go
```diff
@@ -39,14 +100,30 @@ func RunLogin(ctx CommandContext, args LoginArgs) error {
...
 	}
-	err = api.CheckCredentials()
+	err = apiConnection.CheckCredentials()
 	if err != nil {
 		return fmt.Errorf("Unable to contact ScalewayAPI: %s", err)
 	}
+	if args.IndexSSH != "" {
+		uuid, err := apiConnection.GetUUID()
+		if err != nil {
+			return fmt.Errorf("Unable to contact ScalewayAPI: %s", err)
```
moul: Not 100% sure, but I think we should consider than `login` works even if the SSH key failed to upload, so maybe we can just print the error with `logrus.Errorf`, or we can still return the error but maybe we can call  `cfg.Save()` just after the successful `apiConnection.CheckCredentials()`

moul:
LGTM <3

PR #127, opened 2015-08-20
Bump to Golang 1.5

moul:
Not sure travis is already ready for Golang 1.5


moul:
Maybe we should add a check on the makefile when releasing a new version

Issue #130, opened 2015-08-21
`scw info` should print more info from the API


moul:
- [ ] usage
- [x] quotas
- [ ] user/organization information
- [x] ssh keys
- [x] dashboard
- [x] display each APIs URL

QuentinBrosse:
This issue has been automatically closed because it has not had recent activity. Feel free to reopen it if you think that it is necessary.
Note that we are working on a totally new version of the CLI to support all the new coming products.
Thank you for your contributions and stay tuned! :wink:

PR #148, opened 2015-09-01
fix 130 add infos

QuentinPerez:
#130

### DIFF
--- a/pkg/commands/info.go
+++ b/pkg/commands/info.go
@@ -12,6 +12,7 @@ import (
 	"github.com/scaleway/scaleway-cli/vendor/github.com/kardianos/osext"
 
 	"github.com/scaleway/scaleway-cli/pkg/config"
+	"github.com/scaleway/scaleway-cli/pkg/utils"
 )
 
 // InfoArgs are flags for the `RunInfo` function
@@ -41,5 +42,23 @@ func RunInfo(ctx CommandContext, args InfoArgs) error {
 	fmt.Fprintf(ctx.Stdout, "  Snapshots: %d\n", ctx.API.Cache.GetNbSnapshots())
 	fmt.Fprintf(ctx.Stdout, "  Volumes: %d\n", ctx.API.Cache.GetNbVolumes())
 	fmt.Fprintf(ctx.Stdout, "  Bootscripts: %d\n", ctx.API.Cache.GetNbBootscripts())
+	user, err := ctx.API.GetUser()
+	if err != nil {
+		return fmt.Errorf("Unable to get your SSH Keys")
+	} else {
+		if len(user.SSHPublicKeys) == 0 {
+			fmt.Fprintln(ctx.Stdout, "You have no ssh keys")
+		} else {
+			fmt.Fprintln(ctx.Stdout, "SSH Keys:")
+			for id, key := range user.SSHPublicKeys {
+				fingerprint, err := utils.SSHGetFingerprint(key.Key)
+				if err != nil {
+					return err
+				} else {
+					fmt.Fprintf(ctx.Stdout, "  [%d] %s", id, fingerprint)
+				}
+			}
+		}
+	}
 	return nil
 }
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -11,6 +11,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"net"
 	"os"
 	"os/exec"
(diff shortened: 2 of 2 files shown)


moul:
Nice, I need to test it before LGTM, but it seems pretty good

pkg/commands/info.go
```diff
@@ -41,5 +42,23 @@ func RunInfo(ctx CommandContext, args InfoArgs) error {
 	fmt.Fprintf(ctx.Stdout, "  Snapshots: %d\n", ctx.API.Cache.GetNbSnapshots())
 	fmt.Fprintf(ctx.Stdout, "  Volumes: %d\n", ctx.API.Cache.GetNbVolumes())
 	fmt.Fprintf(ctx.Stdout, "  Bootscripts: %d\n", ctx.API.Cache.GetNbBootscripts())
+	user, err := ctx.API.GetUser()
+	if err != nil {
+		fmt.Println("Unable to get your SSH Keys")
```
moul: Use `fmt.Fprintln(ctx.Stdout, ...)` instead of `fmt.Println(...)`

pkg/commands/info.go
```diff
@@ -41,5 +42,23 @@ func RunInfo(ctx CommandContext, args InfoArgs) error {
...
+	} else {
+		if len(user.SSHPublicKeys) == 0 {
+			fmt.Println("You have no ssh keys")
```
moul: Use `fmt.Fprintln(ctx.Stdout, ...)` instead of `fmt.Println(...)`

pkg/commands/info.go
```diff
@@ -41,5 +42,23 @@ func RunInfo(ctx CommandContext, args InfoArgs) error {
...
+		} else {
+			fmt.Println("SSH Keys:")
+			for id, key := range user.SSHPublicKeys {
+				fingerprint, err := utils.SSHGetFingerprint(key.Key)
+				if err != nil {
+					fmt.Println(err)
```
moul: Use `fmt.Fprintln(ctx.Stdout, ...)` instead of `fmt.Println(...)`

pkg/commands/info.go
```diff
@@ -41,5 +42,23 @@ func RunInfo(ctx CommandContext, args InfoArgs) error {
...
+				} else {
+					fmt.Printf("  [%d] %s", id, fingerprint)
```
moul: Use `fmt.Fprintf(ctx.Stdout, ...)` instead of `fmt.Printf(...)`

pkg/commands/info.go
moul: Use `return fmt.Errorf` or `return err`

pkg/commands/info.go
```diff
@@ -41,5 +42,23 @@ func RunInfo(ctx CommandContext, args InfoArgs) error {
...
 	fmt.Fprintf(ctx.Stdout, "  Volumes: %d\n", ctx.API.Cache.GetNbVolumes())
 	fmt.Fprintf(ctx.Stdout, "  Bootscripts: %d\n", ctx.API.Cache.GetNbBootscripts())
+	user, err := ctx.API.GetUser()
+	if err != nil {
+		fmt.Println("Unable to get your SSH Keys")
+	} else {
+		if len(user.SSHPublicKeys) == 0 {
+			fmt.Println("You have no ssh keys")
+		} else {
+			fmt.Println("SSH Keys:")
```
moul: Use `fmt.Fprintln(ctx.Stdout, ...)` instead of `fmt.Println(...)`

pkg/utils/utils.go
```diff
@@ -211,3 +212,25 @@ However, you can access your serial using a web browser:
...
+	buff := []byte(key)
+	bytesWritten := 0
+	for bytesWritten < len(buff) {
+		nb, err := tmp.Write(buff[bytesWritten:])
+		if err != nil {
+			return "", fmt.Errorf("Unable to write: %v", err)
+		}
+		bytesWritten += nb
+	}
+	defer os.Remove(tmp.Name())
```
moul: This defer must be just before the `buff := []byte(key)`

pkg/utils/utils.go
```diff
@@ -211,3 +212,25 @@ However, you can access your serial using a web browser:
...
+	ret, err := exec.Command("ssh-keygen", "-l", "-f", tmp.Name()).Output()
```
moul: Create an issue for later: do not use external program to compute the fingerprint

moul:
LGTM

PR #152, opened 2015-09-02
Reworked the check image script (now checking hostname).

aimxhaisse:


moul:
missing some ascii banner, but LGTM

PR #189, opened 2015-09-22
Various cache file handling fixes

ebfe:
`os.Rename` only works if src and dest are on the same partition. So if `$TMPDIR`
and `$HOME` are on different paritions the cache file was never written and a lot
of temp files were left behind.


aimxhaisse:
Thanks a lot for these fixes!

LGTM

moul:
LGTM

PR #206, opened 2015-10-12
Use new location fields

moul:


QuentinPerez:
Region has been deleted, else LGTM

moul:
Fixed

PR #215, opened 2015-10-16
Use anonuuid to check UUID

QuentinPerez:


moul:
LGTM, but try to split code and party commits

PR #216, opened 2015-10-16
Change uuid check

QuentinPerez:


moul:
ahah :) when we will rewrite and package the cache resolver, we should definitely write unit tests for those things

moul:
LGTM

PR #224, opened 2015-10-19
Fix bug that starting an already started server succeeds.

arianvp:
The code should catch when the server is already started and you try
to call "scw start <id>" but currently that doesn't happen due to a
typo s/!=/==/.


moul:
Good catch, I made this bad copy/paste when migrating from a monolithic app to multiple packages, thanks

PR #246, opened 2015-12-17
Support of `scw run --userdata=...`

### LINKED ISSUE #202: Support of `scw run --userdata=...` option
So we can bootstrap a server with userdata in a unique command, especially cool when using cloud-init

QuentinPerez:
fix #202


moul:
LGTM :sheep:

PR #252, opened 2016-01-04
Fix 151

QuentinPerez:


moul:
Super cool ! do you know if it's always reliable, i.e. does it also works for non RSA keys ?

QuentinPerez:
I tested with the keys that golang can handle https://github.com/golang/crypto/blob/master/ssh/keys.go#L27-L31

moul:
Okay, LGTM

PR #253, opened 2016-01-04
Fix Travis builds

moul:


QuentinPerez:
We need to update the vendor directory for adding `crypto/ssh`

PR #256, opened 2016-01-06
Add region and arch field for images

QuentinPerez:
:warning: Don't merge yet :warning:

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -1170,7 +1170,8 @@ func (s *ScalewayAPI) PostImage(volumeID string, name string, bootscript string,
 	if err = json.Unmarshal(body, &image); err != nil {
 		return "", err
 	}
-	s.Cache.InsertImage(image.Image.Identifier, image.Image.Name)
+	// FIXME region, arch, owner, title
+	s.Cache.InsertImage(image.Image.Identifier, "fr-1", image.Image.Arch, image.Image.Organization, image.Image.Name)
 	return image.Image.Identifier, nil
 }
 
@@ -1280,7 +1281,8 @@ func (s *ScalewayAPI) GetImages() (*[]ScalewayImage, error) {
 		return nil, err
 	}
 	for _, image := range images.Images {
-		s.Cache.InsertImage(image.Identifier, image.Name)
+		// FIXME region, arch, owner, title
+		s.Cache.InsertImage(image.Identifier, "fr-1", image.Arch, image.Organization, image.Name)
 	}
 	return &images.Images, nil
 }
@@ -1302,7 +1304,8 @@ func (s *ScalewayAPI) GetImage(imageID string) (*ScalewayImage, error) {
 	if err = json.Unmarshal(body, &oneImage); err != nil {
 		return nil, err
 	}
-	s.Cache.InsertImage(oneImage.Image.Identifier, oneImage.Image.Name)
+	// FIXME region, arch, owner, title
+	s.Cache.InsertImage(oneImage.Image.Identifier, "fr-1", oneImage.Image.Arch, oneImage.Image.Organization, oneImage.Image.Name)
 	return &oneImage.Image, nil
 }
 
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -1139,7 +1139,8 @@ func (s *ScalewayAPI) PostSnapshot(volumeID string, name string) (string, error)
 	if err = json.Unmarshal(body, &snapshot); err != nil {
 		return "", err
 	}
-	s.Cache.InsertSnapshot(snapshot.Snapshot.Identifier, snapshot.Snapshot.Name)
+	// FIXME region, arch, owner, title
+	s.Cache.InsertSnapshot(snapshot.Snapshot.Identifier, "fr-1", "", snapshot.Snapshot.Organization, snapshot.Snapshot.Name)
 	return snapshot.Snapshot.Identifier, nil
 }
 
@@ -1346,7 +1347,8 @@ func (s *ScalewayAPI) GetSnapshots() (*[]ScalewaySnapshot, error) {
 		return nil, err
 	}
 	for _, snapshot := range snapshots.Snapshots {
-		s.Cache.InsertSnapshot(snapshot.Identifier, snapshot.Name)
+		// FIXME region, arch, owner, title
+		s.Cache.InsertSnapshot(snapshot.Identifier, "fr-1", "", snapshot.Organization, snapshot.Name)
 	}
 	return &snapshots.Snapshots, nil
 }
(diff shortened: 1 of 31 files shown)


pkg/api/cache.go
```diff
@@ -19,22 +19,30 @@ import (
 	"github.com/renstrom/fuzzysearch/fuzzy"
 )
 
+const ( // cache fields
+	REGION   = iota
+	ARCH     = iota
+	OWNER    = iota
+	TITLE    = iota
+	MAXFIELD = iota
+)
```
moul: you don't need to specify iota each time

moul: Prefix the constants

pkg/api/api.go
```diff
@@ -1736,82 +1759,75 @@ func (s *ScalewayAPI) GetDashboard() (*ScalewayDashboard, error) {
...
 	for _, result := range results {
-		fmt.Fprintf(w, "- %s\t%s\t%s\n", result.TruncIdentifier(), result.CodeName(), result.Name)
+		arch = result.Arch
+		if arch == "" {
+			arch = "n/a"
+		}
+		fmt.Fprintf(w, "- %s\t%s\t%s\t%s\n", result.TruncIdentifier(), result.CodeName(), result.Name, arch)
 	}
-	return nil
+	return fmt.Errorf("Too many candidates for %s (%d)", needle, len(results))
```
moul: Do you have an example of the new output with the title *under* the list of candidates ?

pkg/cli/cmd_help.go
```diff
@@ -67,7 +66,7 @@ func runHelp(cmd *Command, rawArgs []string) error {
 				return command.PrintUsage()
 			}
 		}
-		logrus.Fatalf("Unknown help topic `%s`.  Run 'scw help'.", name)
+		return fmt.Errorf("Unknown help topic `%s`.  Run 'scw help'.", name)
```
moul: Due to this, I have ⚠️ on compilation

```
.//pkg/cli/cmd_help.go:76: unreachable code
```

pkg/api/api.go
QuentinPerez:
```console
- 5f19794c          bootscript:3-2-34-apparmor-31          3.2.34-apparmor #31            arm
- 7c7fcac7          bootscript:4-2-5-apparmor-259          4.2.5-apparmor #259            arm
- 1e7a3063          bootscript:armv7l-4-3-3-std-1-latest   armv7l 4.3.3 std #1 (latest)   arm
FATA[0000] cannot execute 'run': failed to create server: Too many candidates for m (3)
```

pkg/cli/cmd_help.go
```diff
@@ -67,7 +66,7 @@ func runHelp(cmd *Command, rawArgs []string) error {
...
 	} else {
```
moul: You can remove this "else" completely

pkg/api/api.go
```diff
@@ -1736,82 +1759,75 @@ func (s *ScalewayAPI) GetDashboard() (*ScalewayDashboard, error) {
...
 
 // GetServerID returns exactly one server matching or dies
-func (s *ScalewayAPI) GetServerID(needle string) string {
+func (s *ScalewayAPI) GetServerID(needle string) (string, error) {
 	// Parses optional type prefix, i.e: "server:name" -> "name"
 	_, needle = parseNeedle(needle)
 
 	servers, err := s.ResolveServer(needle)
 	if err != nil {
-		log.Fatalf("Unable to resolve server %s: %s", needle, err)
```
moul: From what I remembert, we used `log.Fatalf` here instead of standard error return, because GetServerID may be called in a  goroutine, we need to check that the goroutine will forward the error to a channel or at least move the `fatal` behaviour inside the goroutine, so we can keep this helper fatal-proof as you just did

pkg/api/api.go
```diff
@@ -1736,82 +1759,75 @@ func (s *ScalewayAPI) GetDashboard() (*ScalewayDashboard, error) {
...
 	_, needle = parseNeedle(needle)
 
 	images, err := s.ResolveImage(needle)
 	if err != nil {
-		log.Fatalf("Unable to resolve image %s: %s", needle, err)
+		return "", "", fmt.Errorf("Unable to resolve image %s: %s", needle, err)
 	}
 	if len(images) == 1 {
-		return images[0].Identifier
+		return images[0].Identifier, images[0].Arch, nil
```
moul: I think that you will soon need to add the region here, so maybe we should switch to a structure

pkg/api/cache.go
```diff
@@ -19,22 +19,35 @@ import (
...
+	REGION = iota
+	// ARCH permits to access at the arch field
+	ARCH
+	// OWNER permits to access at the owner field
+	OWNER
+	// TITLE permits to access at the title field
+	TITLE
+	// MAXFIELD is used to determine the size of array
+	MAXFIELD
+)
```
moul: Prefix the constants

pkg/api/cache.go
```diff
@@ -117,6 +141,18 @@ func (s *ScalewayResolverResult) CodeName() string {
 	return fmt.Sprintf("%s:%s", strings.ToLower(identifierTypeName(s.Type)), name)
 }
 
+// FilterByArch deletes the elements which not match with arch
+func (s *ScalewayResolverResults) FilterByArch(arch string) {
+REDO:
```
moul: newbie question, are the uppercase labels also exported ?

pkg/api/helpers.go
```diff
@@ -229,7 +228,7 @@ func InspectIdentifiers(api *ScalewayAPI, ci chan ScalewayResolvedIdentifier, cj
 			if len(idents.Identifiers) == 0 {
 				log.Errorf("Unable to resolve identifier %s", idents.Needle)
 			} else {
-				showResolverResults(idents.Needle, idents.Identifiers)
+				logrus.Fatal(showResolverResults(idents.Needle, idents.Identifiers))
```
moul: oh nice, you actually moved the logrus.fatal inside the goroutine :) ignore my last comment about this

pkg/api/helpers.go
```diff
@@ -349,12 +348,10 @@ func CreateServer(api *ScalewayAPI, c *ConfigCreateServer) (string, error) {
 			server.Volumes[volumeIDx] = *volumeID
 		}
 	}
-	server.Name = c.Name
-	if c.Bootscript != "" {
-		bootscript := api.GetBootscriptID(c.Bootscript)
-		server.Bootscript = &bootscript
-	}
+	// FIXME build images only on ARM ?
```
moul: typo ?

pkg/commands/images.go
```diff
@@ -145,21 +161,19 @@ func RunImages(ctx CommandContext, args ImagesArgs) error {
...
+			entries = append(entries, entry)
 		}
 	}
-
+	select {
+	case err := <-errChan:
+		return err
+	default:
+		break
+	}
```
moul: nice

pkg/api/api.go
moul: perfect

moul:
LGTM, you can fix my comments now or later as they are clearly not important

We need to test this version for a while so we can adjust the details, so as soon as the PR is merged, I will create a pre-release

pkg/api/helpers.go
moul: ok got it, it is related to `scw run 1G` which will create a new volume

pkg/api/cache.go
QuentinPerez: Good question :), I tested and no, it's not exported

moul: ok, so no problem

PR #261, opened 2016-01-11
Fix #194 1/2

### LINKED ISSUE #194: When connecting using SSH, check if host keys are available in the userdata
Depends on: https://github.com/scaleway/image-tools/issues/36

If the host keys are available you can : 

- [ ] display fingerprints
- [ ] register fingerprint on the local host (ssh-keygen) + remove this: https://github.com/scaleway/scaleway-cli/blob/master/pkg/utils/utils.go#L83

I suggest you to begin with 1 as the userdata API is still very young and most of the actual images won't have scripts that generates the fingerprints

QuentinPerez:
fix #194

### DIFF
--- a/README.md
+++ b/README.md
@@ -1154,7 +1154,8 @@ $ scw inspect myserver | jq '.[0].public_ip.address'
 
 ### master (unreleased)
 
-# Add a warn message when using `ssh exec` on host without public ip nor gateway ([#171](https://github.com/scaleway/scaleway-cli/issues/171))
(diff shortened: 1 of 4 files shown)


moul:
Needs a rebase

QuentinPerez:
Done :)

moul:
I think we will need a message before printing out all the keys, but I have no idea for now, so lets merge it and find by using this new feature

moul:
LGTM

PR #271, opened 2016-01-28
Update README.md

QuentinPerez:


README.md
```diff
@@ -67,10 +67,12 @@ For node version, check out [scaleway-cli-node](https://github.com/moul/scaleway
 
 ## Setup
 
+:warning: Ensure you have a go version `>= 1.5`
```
moul: this is only if you use the `go get`, so move this line just above the `go get` line

moul:
LGTM

PR #283, opened 2016-02-21
Enable creation of public IP-less servers via special IP address names

ElNounch:


moul:
Thank you @ElNounch, it looks good to me


@QuentinPerez, can you give a try tomorrow to see if every cases are handled ?

We will also need to:

* Update the Changeloga
* Add usage examples in `README.md`
* Add usage examples in `scw create -h`
* Port this to `scw run` too

QuentinPerez:
Tested and LGTM 😃

ElNounch:
Commited port to `scw run` in what should be a retro-compatible way.

QuentinPerez:
Thank you @ElNounch 😉

moul:
@QuentinPerez, can you update the Changelog ?

PR #309, opened 2016-03-16
Fix wget with http_proxy on busybox x86_64

moul:
On busybox x86_64, busybox's wget is compiled with http proxy support but without wgetrc support

This commit fills an environment file instead of writing a wgetrc file


QuentinPerez:
LGTM

Issue #334, opened 2016-04-06
IPv6 fields


ewoutp:
Please add the IPv6 related fields to the server definitions (in the api package).

moul:
Fixed with #335, not yet released, but you can build the latest version by yourself

ewoutp:
This solves only one side of it. The creation of servers.
I also need to read the IPv6 address in a "get-server" or "list-servers" action.

QuentinPerez:
Hi @ewoutp, the PR #339 solves your needs ?

ewoutp:
Yes, looks good. thx

PR #335, opened 2016-04-06
Ipv6

QuentinPerez:
see #334


moul:
LGTM

moul:
I think we should enable it by default :)

Issue #337, opened 2016-04-06
unexpected end of JSON input


Hades32:
OS: Windows 10
Shell: Msys2, Part of Git 2.7.0.windows.1
Logged in with token and organisation.

marti@WIN-SJ0A7GO32DV  /d/Temp
$ scw images
time="2016-04-06T23:26:42+02:00" level=fatal msg="cannot execute 'images': unable to fetch images from the Scaleway API: unexpected end of JSON input"

QuentinPerez:
Hi @Hades32, thank you for your feedback ☺️

Do you have the latest version of scw ?

Hades32:
Ah sorry, of course THAT I forgot :)

I downloaded the newest release 1.9

moul:
@Hades32 can you run `scw -D images` (`-D` is for debug)

Hades32:
Hmm, looks like I was not supposed to enter the name of the organisation...

$ scw -D images
time="2016-04-07T22:28:48+02:00" level=debug msg="curl -X GET -H \"Content-Type: application/json\" -H \"User-Agent: scw/v1.9.0\" -H \"X-Auth-Token: 00000000-0000-1000-0000-000000000000\" https://api-marketplace.scaleway.com/images/"
time="2016-04-07T22:28:49+02:00" level=debug msg="curl -X GET -H \"Content-Type: application/json\" -H \"User-Agent: scw/v1.9.0\" -H \"X-Auth-Token: 00000000-0000-1000-0000-000000000000\" \"https://api.scaleway.com/images?organization=invaliduuid\""
time="2016-04-07T22:28:50+02:00" level=debug msg="Validation Error" Message= StatusCode=400 Type="invalid_request_error"
time="2016-04-07T22:28:50+02:00" level=debug msg="  organization:                  [Martin Rauscher is not a valid UUID.]"
time="2016-04-07T22:28:50+02:00" level=fatal msg="cannot execute 'images': unable to fetch images from the Scaleway API: unexpected end of JSON input"

Hades32:
So I guess the bug is just a bad error message...

QuentinPerez:
Today, I tested scw on windows and it doesn't work as excepted 😑, I will try to investigate and fix the compatibility

moul:
#343 should (at least partially) fix this

QuentinPerez:
[...]

PR #348, opened 2016-04-15
little patches

QuentinPerez:
- fix comment
- fix bad error message #337 
- fix commercialtype check
- use `go list`


moul:
LGTM

PR #363, opened 2016-06-02
Use proxy configuration from environment in 'scw attach' command

debovema:
This will use HTTP_PROXY and HTTPS_PROXY environment variables in 'scw attach' command.


QuentinPerez:
LGTM thank you 😊

moul:
Thank you @debovema

PR #367, opened 2016-06-16
Add ipv6

QuentinPerez:
`scw [create | run] --ipv6`


moul:
LGTM, we should also add away to patch an existing server

PR #368, opened 2016-06-20
Remove http2curl dependency

nicolai86:
Instead of printing http requests to play around with, the API SDK should be easy to use by itself (expose client use cases) & lightweight so it itself can be easily vendored into 3rd party tools (e.g. terraform). 

For now,  I'd like to remove seemingly unnecessary dependencies (specifically logrus, http2curl & anonuuid), to ease getting my [terraform provider for scaleway](https://github.com/nicolai86/terraform/commits/feature/scaleway) merged.

Looking at the AWS SDK itself, introducing a customer logger interface looks like the clean solution, which would allow us to re-introduced the http2curl logging functionality by allowing users of the SDK to inject a custom logger, which has functions to log http requests specifically; the user can then decide to use whatever mechanism he sees fit, which would allow us to push this dependency to the user, which should be preferred for an API SDK.

In this PR I'm just removing the http2curl dependency, without introducing a custom logger interface, to keep the PR small. If you like the idea I'll go ahead to get a proper logger interface setup.
The logger interface is required anyway to get logrus & anonuuid out of the picture, but that's a separate discussion.


moul:
👍 for creating logger interfaces and remove all the unnecessary dependencies for the API clients users

but we need to keep anonuuid and http2curl in the cli so we can continue to ask our users some outputs for support

moul:
@QuentinPerez and I would be happy to help you on this point, do not hesitate if you feel it will take too much of your time

nicolai86:
@moul I understand that use case. Introducing a pluggable Logger would allow us to keep the current logging as is, while also being able to use the SDK without anonuuid, logrus and http2curl. 

I'll open another PR which will introduce the Logger interface. thanks for the feedback.

moul:
Perfect, looking forward for your pull-request

PR #369, opened 2016-06-20
Introduce Logger

nicolai86:
Following #368 this PR introduces a `Logger` interface which allows us to extract the `http2curl` and `logrus` dependencies into the `cli` & `command` packages.

The `Logger` contains two functions - `LogHTTP` to handle `http.Request`, and `Log` for plain `string`.
The `cli` configures the `ScalewayAPI` instance to retain the old logging behaviour. By default, the logging is not very verbose, logging only messages & API endpoints being used.

Also, I've replaced `anonuuid` with static, fake uuids. Not sure why we need "randomized" uuids if the only concern is to sanitize logs from real credentials.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1184,6 +1184,7 @@ $ scw inspect myserver | jq '.[0].public_ip.address'
 
 ### master (unreleased)
 
+* gotty-client enable ProxyFromEnviromnent ([#363](https://github.com/scaleway/scaleway-cli/pull/363)) (@debovema)
 * `scw inspect` fix panic ([#353](https://github.com/scaleway/scaleway-cli/issues/353))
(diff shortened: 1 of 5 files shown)


nicolai86:
@QuentinPerez thanks for improving the PR!

QuentinPerez:
LGTM, ping @moul 😊

moul:
LGTM, thank you

moul:
Well done guys

moul:
@QuentinPerez can you credit @nicolai86 in the changelog ?

nicolai86:
Thanks for being so responsive! I'll continue my quest to get scaleway into terraform, then. 🙏🏼

nicolai86:
For those interested in the terraform provider, you can take a look here: https://github.com/hashicorp/terraform/pull/7331

PR #372, opened 2016-06-25
Remove single logrus invocation from cache

nicolai86:
The cache is already using `log` in different places, so replacing this single invocation with `log` instead of `logrus`.


moul:
LGTM

PR #377, opened 2016-07-12
Fix 371

### LINKED ISSUE #371: Greeting message when login through the CLI?
It could be nice to have a motd or something like that when using `scw login`.

Actually it's just asking you if you want to upload an SSH key, and doesn't tell you if you are properly logged in.

QuentinPerez:
closes #371 

This PR adds a motd.

```console
Login (cloud.scaleway.com): qperez@ocs.online.net
Password:
Do you want to upload an SSH key ?
[0] I don't want to upload a key !
[1] github.pub
[2] id_rsa.pub
Which [id]: 0

You are now authenticated on Scaleway.com as Quentin Perez.
You can list your existing servers using `scw ps` or create a new one using `scw run ubuntu-xenial`.
You can get a list of all available commands using `scw -h` and get more usage examples on github.com/scaleway/scaleway-cli.
Happy cloud riding.
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -1184,6 +1184,7 @@ $ scw inspect myserver | jq '.[0].public_ip.address'
 
 ### master (unreleased)
 
+* `scw login` Add motd when you are already logged ([#371](https://github.com/scaleway/scaleway-cli/issues/371))
 * Fix error message with `--commercial-type=c2m` ([#374](https://github.com/scaleway/scaleway-cli/issues/374))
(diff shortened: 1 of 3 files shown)


QuentinPerez:
@nicolai86 I had a new logger that displays nothing here, maybe you should use it too, to disable the logging in the terraform test ?

nicolai86:
@quentinperez I'll add a logger that conforms to terraform logging practice later today. That way you can opt into seeing output if necessary

PR #378, opened 2016-07-12
Add DetachIP helper

nicolai86:
While working on my [terraform provider](https://github.com/hashicorp/terraform/pull/7331) I had the need to detach IPs from servers.

This doesn't seem to be exposed in the SDK so I added a [helper function](https://github.com/hashicorp/terraform/pull/7331/commits/51d188233ade206f3b232b090f83b985cc0bcfe0). However, I think it should be public within the SDK.

One question tho: do we need to stop servers in order to detach IPs ?

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -2313,6 +2313,35 @@ func (s *ScalewayAPI) AttachIP(ipID, serverID string) error {
 	return err
 }
 
+// DetachIP detaches an IP from a server
+func (s *ScalewayAPI) DetachIP(ipID string) error {
+	var update struct {
+		Address      string `json:"address"`
+		ID           string `json:"id"`
+		Organization string `json:"organization"`
+	}
+
+	ip, err := s.GetIP(ipID)
+	if err != nil {
+		return err
+	}
+
+	if ip.IP.Server.Identifier == "" {
+		return nil
+	}
+
+	update.Address = ip.IP.Address
+	update.ID = ip.IP.ID
+	update.Organization = ip.IP.Organization
+	resp, err := s.PutResponse(ComputeAPI, fmt.Sprintf("ips/%s", ipID), update)
+	if err != nil {
+		return err
+	}
+	defer resp.Body.Close()
+	_, err = s.handleHTTPError([]int{200}, resp)
+	return err
+}
+
 // DeleteIP deletes an IP
 func (s *ScalewayAPI) DeleteIP(ipID string) error {
 	resp, err := s.DeleteResponse(ComputeAPI, fmt.Sprintf("ips/%s", ipID))


pkg/api/api.go
```diff
@@ -2313,6 +2313,35 @@ func (s *ScalewayAPI) AttachIP(ipID, serverID string) error {
 	return err
 }
 
+// DetachIP attachs an IP to a server
```
QuentinPerez: DetachIP detachs an IP from a server

pkg/api/api.go
```diff
@@ -2313,6 +2313,35 @@ func (s *ScalewayAPI) AttachIP(ipID, serverID string) error {
...
+func (s *ScalewayAPI) DetachIP(ipID, serverID string) error {
```
QuentinPerez: You don’t need the ServerID

pkg/api/api.go
```diff
@@ -2313,6 +2313,35 @@ func (s *ScalewayAPI) AttachIP(ipID, serverID string) error {
...
+	var update struct {
+		Address      string  `json:"address"`
+		ID           string  `json:"id"`
+		Reverse      *string `json:"reverse"`
```
QuentinPerez: I don’t think that you need the reverse field

pkg/api/api.go
```diff
@@ -2313,6 +2313,35 @@ func (s *ScalewayAPI) AttachIP(ipID, serverID string) error {
...
+	}
+
+	if ip.IP.Server.Identifier == "" {
+		return nil
+	}
+
+	update.Address = ip.IP.Address
+	update.ID = ip.IP.ID
+	update.Organization = ip.IP.Organization
+	resp, err := s.PutResponse(ComputeAPI, fmt.Sprintf("ips/%s", ipID), update)
```
QuentinPerez: Can you add this code below
```console
if resp != nil {
	defer resp.Body.Close()
}
```

QuentinPerez:
Hi @nicolai86, one more time, thank you for your work ☺️

Can you take a look on my comments ?

nicolai86:
Late night copy pasta and this is the result ~.~
Thanks for taking a look I'll clean this up today.

pkg/api/api.go
nicolai86: Done; see diff

nicolai86:
@QuentinPerez thanks for the quick review yesterday. I've cleaned up the PR accordingly.

pkg/api/api.go
```diff
@@ -2313,6 +2313,35 @@ func (s *ScalewayAPI) AttachIP(ipID, serverID string) error {
...
+	if err != nil {
+		return err
+	}
+	defer resp.Body.Close()
```
QuentinPerez: You have to add this code between `s.PutResponse` and `if err != nil`

```golang
if resp != nil {
    defer resp.Body.Close()
}
```

according to  http://devs.cloudimmunity.com/gotchas-and-common-mistakes-in-go-golang/#anameclose_http_resp_bodyaclosinghttpresponsebody

QuentinPerez:
I merge your PR, and I will fix some issues on our `ScalewayIPScaleway` to detach an IP more easily.

Thanks a lot !!! 😊 👍🏻

PR #379, opened 2016-07-13
ips

QuentinPerez:
`scw _ips` : add --detach flag
Change the `ScalewayIPDefinition` fields to detach an IP easily #378  @nicolai86


moul:
LGTM

Issue #254, opened 2016-01-05
Do not wait hardcoded 30 seconds when `scw exec --gateway`


moul:
```console
➜  ~  scw -D exec -w --gateway=gatewesh suspicious-hopper
DEBU[0000] curl -X GET -H "X-Auth-Token: 00000000-0000-1000-0000-000000000000" -H "Content-Type: application/json" https://api.scaleway.com/servers/f9e9bfa4-87ef-430d-aa29-813c1d0491c3
DEBU[0000] The server will be accessed using the gateway '212.47.228.0' as a SSH relay
DEBU[0000] Waiting for server to be ready
DEBU[0000] curl -X GET -H "Content-Type: application/json" -H "X-Auth-Token: 00000000-0000-1000-0000-000000000000" https://api.scaleway.com/servers/cf872fe2-bfca-4d83-a91d-248dcbe05767
INFO[0001] Server changed state to 'running'
DEBU[0001] Waiting for gateway SSH port
DEBU[0001] Waiting 30 more seconds, for SSH to be ready
```

The 30 secs hardcoded wait is only necessary when the target server is booting, but when running `scw exec -w`, the server may already be booted, so this hardcoded wait should only apply on `scw run` and not on `scw exec`

Issue #157, opened 2015-09-04
Show DNS names when possible


moul:
- [x] add DNS urls in `scw inspect`
- [x] add DNS urls when creating/runing/execing a server if DEBUG mode is enabled

See: https://community.scaleway.com/t/expose-server-names-under-dns/426/5?u=manfred

PR #399, opened 2016-09-03
Add --user {cp,exec,kill,port,run,top}

### LINKED ISSUE #396: SSH user should be configurable
When disabling root ssh login for security concerns, we should be able to set another ssh user to connect.

QuentinPerez:
closes #396 #157 #254


moul:
LGTM

zifeo:
Thanks!

PR #411, opened 2016-09-28
Fix #410: fix detection of paginated response

### LINKED ISSUE #410: 'scw login' - "Unable to find your organization"
Hello!

When I do a `scw login` after typing my credentials I've got this error:
```
FATA[0010] cannot execute 'login': Unable to find your organization
```
I've dug a bit and I saw that when I try the api itself with the url `https://account.scaleway.com/organizations`, I've got a nice 200 OK with 1 result and header `X-Total-Count: 1`.

That means I have an organization. So I've dug further, into the code this time and it appears that in the 
`GetResponsePaginate` func in `pkg/api/api.go` there is a problem with the handling of whether a response must be paginated or not.

```go
// line 931
count := ""
if count = resp.Header.Get("X-Total-Count"); count == "" { // not paginated
    if len(values) == 0 {
        return s.response("GET", fmt.Sprintf("%s/%s", strings.TrimRight(apiURL, "/"), resource), nil)
    }
    return s.response("GET", fmt.Sprintf("%s/%s?%s", strings.TrimRight(apiURL, "/"), resource, values.Encode()), nil)
}
```

In my case `count` is `"1"`, because I have one organization, but it will be considerd like a paginated response (see the code above)

A paginated response should be detected with calculating the number of pages, here:
```
count / perPage = page
1 / 50 = 0.02 // (less than a page)
```

Thanks for your work, I'll bring you a fix in a PR

devster:
This is a fix for the issue #410

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -927,22 +927,31 @@ func (s *ScalewayAPI) GetResponsePaginate(apiURL, resource string, values url.Va
 	if err != nil {
 		return nil, err
 	}
-	count := ""
-	if count = resp.Header.Get("X-Total-Count"); count == "" { // not paginated
+
+	count := resp.Header.Get("X-Total-Count")
+	var maxElem int
+	if count == "" {
+		maxElem = 0
+	} else {
+		maxElem, err = strconv.Atoi(count)
+		if err != nil {
+			return nil, err
+		}
+	}
+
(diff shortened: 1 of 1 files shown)


QuentinPerez:
LGTM thank you for you PR really appreciate ! ☺️

PR #423, opened 2016-10-25
Fix volume size

QuentinPerez:
fix unmarshal error

```console
FATA[0002] cannot execute 'images': unable to fetch volumes from the Scaleway API: json: cannot unmarshal number 5e+10 into Go value of type uint64
```


moul:
LGTM

PR #440, opened 2016-12-07
chore/cleanup: assume response body is present

nicolai86:
to quote the [Golang documentation for http.Client Do](https://godoc.org/net/http#Client.Do):

```
If the returned error is nil, the Response will contain a non-nil Body which the
user is expected to close.
```

we can simplify the current http client usage quite a lot by just moving `defer
Close()` down some lines.

### change summary:

before:

```
resp, err := …
if resp != nil {
    defer resp.Body.Close()
}
if err != nil {
    return err
}
```

after

```
resp, err := …
if err != nil {
    return err
}
defer resp.Body.Close()
```

this does not change any program semantics. the change just cleans up the code a little.


QuentinPerez:
Thanks @nicolai86

Issue #373, opened 2016-06-25
Discussion: why have an client side API cache?


nicolai86:
I'm wondering why the API SDK has a local cache? Assuming you're already caching on the server side the client cache probably doesn't add a lot of benefit but some maintenance burden.

Removing the cache would simplify the API SDK a little, and also remove one possible point for errors (less caches = less stale results)

moul:
The cache is used to avoid listing resources each time we need to get a resourced based on something else than its ID, for example querying a server based on its hostname

```console
$ scw _flush-cache >/dev/null
$ time scw --debug inspect --format '{{ .Hostname }}' server:swarm-node-1

DEBU[0000] Filling the cache
DEBU[0000] curl -X GET -H "Content-Type: application/json" -H "User-Agent: scw/v1.9.0+dev" -H "X-Auth-Token: 00000000-0000-1000-0000-000000000000" https://api.scaleway.com/servers
DEBU[0001] curl -X GET -H "Content-Type: application/json" -H "User-Agent: scw/v1.9.0+dev" -H "X-Auth-Token: 00000000-0000-1000-0000-000000000000" https://api.scaleway.com/servers/22baa68d-e439-48d1-9437-b1a764f34acc
swarm-node-1

0.11s user 0.03s system 8% cpu 1.821 total
$ time scw --debug inspect --format '{{ .Hostname }}' server:swarm-node-1

DEBU[0000] curl -X GET -H "Content-Type: application/json" -H "User-Agent: scw/v1.9.0+dev" -H "X-Auth-Token: 00000000-0000-1000-0000-000000000000" https://api.scaleway.com/servers/22baa68d-e439-48d1-9437-b1a764f34acc
swarm-node-1

0.12s user 0.02s system 29% cpu 0.460 total
```

* First call: 1.8sec
  * listing all servers
  * inspecting the one matching the hostname
* Second call: 0.5sec
  * inspecting the server matching the hostname in the cache

When you create a server with a named `bootscript`, a named `image`, a raw `ip address` and secondary volumes, you can quickly gain several seconds

It is also used by the autocomplete scripts of the CLI that will only look in this cache file to avoid querying the API just for autocomplete

---

As you did in #369, I'm fully 👍 to keep this feature in the CLI, but make it optional when using the API sdk

Looks good to you ?

nicolai86:
@moul thanks for explaining! I fully understand the use case for the CLI. I'll check if there's a nice way to migrate the caching for cli only.

PR #441, opened 2016-12-07
[WIP] chore/caching: optional caching

nicolai86:
In order to follow up on #373 I'm opening a WIP PR to gather feedback as early as possible.

Roughly I'm planning on introducing interfaces for specific API concerns, so we can transparently hook up caching. This way other projects (e.g. terraform) can use the API without caching, and the official scaleway CLI can still use caching. The endresult might look like this:

```
# any cli
api := NewScalewayAPI(…)
ctx.API = CacheDecorator{api}

# api/server.go
type ServerAPI interface {
  GetServers(…) 
  GetServer(…)
  …
}

# api.go
type ScalewayAPI interface {
  ServerAPI
  VolumeAPI
  ImageAPI
  …
}

# cli/cache.go
type CacheDecorator struct {
  api *ScalewayAPI
  cache *ScalewayCache
}

func (d *CacheDecorator) GetServers(…) {
  if servers, ok := cache.ResolveServers() {
     return servers, nil
  }
  servers, _ := d.api.GetServers()
  // update cache
  return servers, nil
}
```

To achive this I'm planning on the following steps:

- [x] split up api into multiple files. The API is really huge and hard to grasp. To enable the following changes I need to structure individual areas better, and splitting the API up by concern should enable me to do just that
- [x] introduce interfaces for specific concerns. This might look like `ServerAPI`, `ImageAPI`, `VolumeAPI`, etc. 
- [ ] introduce decorator struct which passes requests to api /w handled caching
- [ ] modify CLI to use caching decorator around API, remove caching from API implementation

any feedback is welcome


nicolai86:
Closing for now - I don't think patching it like this is the right approach after all. The SDK is tightly coupled to caching in multiple places, which makes the process of making it optional  very  painful.
I think it would be easier if the API client was generated from a definition like swagger, so one can more easily compose the SDK from specific functionalities, but that's multiple big PRs in itself.

PR #457, opened 2017-05-03
Speedup Makefile and gobuild

moul:


brmzkw:
Do you have an idea why the build still fails?

moul:
Yep, it looks like to be `make test` that were not yet ported to the new system

moul:
Fixed:)

brmzkw:
Nice! thanks!

PR #403, opened 2016-09-09
Better handling of secondary volumes size

QuentinPerez:
closes #361  #360

Issue #361, opened 2016-05-14
Better handling of secondary volumes size


moul:
Some new offers have size limits, we should as often as possible create a secondary volume if not manually filled. This is the default behavior on the console for instance.

See https://github.com/scaleway/scaleway-cli/issues/360 and https://github.com/scaleway/docker-machine-driver-scaleway/issues/40

ViViDboarder:
This causes issues with image builder if you have a base image > 50GB.

https://github.com/scaleway/image-builder/issues/14

It would be a lot more complicated, but it would be better, if possible to detect the size of the primary image first before adding in a secondary.

Easy repro steps:
```
root@scw-b6f9a0:~/image-ubuntu# scw -D run --commercial-type="VC1M" 100G
INFO[0000] Server creation ...
DEBU[0000] This server needs a least 50G
DEBU[0000] curl -X POST -d "{\"name\":\"50G\",\"size\":50000000000,\"volume_type\":\"l_ssd\",\"organization\":\"<org>\"}" -H "Content-Type: application/json" -H "User-Agent: scw/v1.10.1" -H "X-Auth-Token: <token>" 'https://api.scaleway.com/volumes'
DEBU[0000] [POST]: https://api.scaleway.com/volumes
DEBU[0002] [Response]: [201]
{"volume": {"size": 50000000000, "name": "50G", "modification_date": "2017-05-23T21:44:48.188970+00:00", "organization": "<orgid>", "export_uri": null, "creation_date": "2017-05-23T21:44:48.188970+00:00", "id": "4f4badba-6bd6-41e3-9ed6-c1c6da6a081b", "volume_type": "l_ssd", "server": null}}
DEBU[0002] curl -X POST -d "{\"name\":\"100G\",\"size\":100000000000,\"volume_type\":\"l_ssd\",\"organization\":\"<org>\"}" -H "Content-Type: application/json" -H "User-Agent: scw/v1.10.1" -H "X-Auth-Token: <token>" 'https://api.scaleway.com/volumes'
DEBU[0002] [POST]: https://api.scaleway.com/volumes
DEBU[0002] [Response]: [201]
[...]

PR #462, opened 2017-05-23
Add detection of main volume size before creating additional volumes

ViViDboarder:
With the current implementation the api will force the creation of
additional volumes to meet the minimum per instance type regardless of
the primary volume size. This causes an issue if the primary volume is
larger than 50G

This patch will compare the primary volume to the min sizes

This expands upon #403 provides a slightly more flexible fix for #361 that also resolves scaleway/image-builder#14

### DIFF
--- a/pkg/api/helpers.go
+++ b/pkg/api/helpers.go
@@ -336,45 +336,6 @@ func CreateServer(api *ScalewayAPI, c *ConfigCreateServer) (string, error) {
 	if c.Env != "" {
 		server.Tags = strings.Split(c.Env, " ")
 	}
-	switch c.CommercialType {
-	case "VC1M", "X64-4GB", "ARM64-4GB":
-		if c.AdditionalVolumes == "" {
-			c.AdditionalVolumes = "50G"
-			log.Debugf("This server needs a least 50G")
-		}
-	case "VC1L", "X64-8GB", "X64-15GB", "ARM64-8GB":
-		if c.AdditionalVolumes == "" {
-			c.AdditionalVolumes = "150G"
-			log.Debugf("This server needs a least 150G")
-		}
-	case "X64-30GB":
-		if c.AdditionalVolumes == "" {
-			c.AdditionalVolumes = "100G 150G"
-			log.Debugf("This server needs a least 300G")
-		}
-	case "X64-60GB":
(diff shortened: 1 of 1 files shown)


brmzkw:
Hi,

Thanks a lot for your pull request.

At Scaleway, we internally maintain a JSON file which describes our offers. It is available [here as a Github gist](https://git.io/vH6Uk). It will probably be exposed by our APIs in a near future.

This JSON file contains:

* the commercial names of our offers : the top level key in the dict. Some products also have alternative names. For instance, "VC1S" describes the same server than "X64-2GB".
* some hardware details: number of CPUs, RAM, NICs, ...
* and more importantly in the context of this PR: the volumes constraints, ie. the minimal/maximal volumes sizes to create a server of a given type.

Don't you think we should go further, and:

* add this JSON file in the scaleway-cli repository
* someway parse and use it to create servers

?

I'm not a golang expert, so I'm not sure of the best way to integrate such a file though.

ViViDboarder:
Oh, that would be very nice!

Golang has some great built in ways to marshal and unmarshal JSON, so parsing this would be very easy. It then seems easy enough to use the returned map to get the volume constraints.

My only concern with doing it here would be that it maybe confusing if it's not used everywhere. Also, the compiling the JSON into the binary might be non-trivial.

Given that... if there was an API available, it might be easier/better to query the restrictions and offerings from that API. That would allow adding and removing of offerings from your server-side without client changes.

brmzkw:
Here we are, I updated our compute APIs:

* https://cp-par1.scaleway.com/products/servers
* https://cp-ams1.scaleway.com/products/servers

Note the C1 servers are not available in Amsterdam.

If you want to include these new APIs in scaleway-cli, I'd be glad to accept the pull request. Otherwise I'll probably work on it sooner or later.

brmzkw:
@tbillon : I think this PR can be closed, isn't it? Correct me if I'm wrong, but I think we only have to release scaleway-cli.

tbillon:
Yes it can probably be closed. My previous commit should do the trick.

brmzkw:
But since @tbillon implemented a dynamic version to find the volumes to create for a given instance, we will go with his version instead.

Again, thanks a lot for your contribution @ViViDboarder

PR #469, opened 2017-07-05
create-image-from-http.sh: use the latest bootscript version

multun:


examples/create-image-from-http.sh
```diff
@@ -16,7 +16,7 @@ fi
 NAME=$(basename "${URL}")
 SNAPSHOT_NAME=${NAME%.*}-$(date +%Y-%m-%d_%H:%M)
 IMAGE_NAME=${IMAGE_NAME:-$SNAPSHOT_NAME}
-IMAGE_BOOTSCRIPT=${IMAGE_BOOTSCRIPT:stable}
+IMAGE_BOOTSCRIPT=${IMAGE_BOOTSCRIPT:latest}
```
bchatelard: Why not leaving the default value to 'empty' ?
No bootscript will use the default one.

PR #486, opened 2018-01-06
Make WaitForServerReady use IPv6 if IPv4 unavailable

meyskens:
This makes WaitForServerReady work for servers that only need an IPv6 address.


brmzkw:
Thanks!

PR #489, opened 2018-02-11
Fix: replace the old bash completion with a new one

africanzoe:
The old bash completion script was buggy and contains options that either don't exist or don't work.


brmzkw:
😍

Issue #446, opened 2017-02-11
region option doesn't work


metal3d:
Hi, it seems that "region" option doesn't work
```
$ scw --version
scw version v1.11.1+dev, build
$ scw ps
... gives par1 server list

$ scw --region=ams1 ps
... gives par1 server list too
```

QuentinPerez:
Hi @metal3d 

`scw ps` lists all servers, it doesn't use global flag `region`, but you can use `./scw  ps -f zone=par1`

metal3d:
@QuentinPerez  Yeah, I tried too:

```
scw ps -a -f zone=ams1
```
The server is down, I should see it, right ? (It doesn't for now)

metal3d:
BTW it seems to be "not so far to work for me" because I can list zone=par1 servers without to see the amsterdam server. The only problem I see is that server that are "down" are not listed if I set zone=ams1.

raarts:
It seems counterintuitive that the `ps` command ignores the `--region` flag. Why is that?

QuentinBrosse:
This issue has been automatically closed because it has not had recent activity. Feel free to reopen it if you think that it is necessary.
Note that we are working on a totally new version of the CLI to support all the new coming products.
Thank you for your contributions and stay tuned! :wink:

PR #496, opened 2018-04-17
WIP: Allow region filtering

sieben:
This is an answer to https://github.com/scaleway/scaleway-cli/issues/446
This PR adds an example showing `scw ps` with a filter.

Should we use the --region flag passed before the command `ps` to filter the API on which the `ps` commands run upon?

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -1208,7 +1208,7 @@ func (s *ScalewayAPI) GetServers(all bool, limit int) (*[]ScalewayServer, error)
 		}
 	)
 
-	serverChan := make(chan ScalewayServers, 2)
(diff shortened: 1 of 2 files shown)


sieben:
@moul What about the --region question? It seems that this questions come back often "region" tag issues.

moul:
> Should we use the --region flag passed before the command ps to filter the API on which the ps commands run upon?

By default, I'm more for using this as a global variable as its (or will be) useful for the majority of the commands in `scw.`

But from what I recently saw, it's more and more standard to put every argument at the end of the command, probably "more POSIX," i.e., https://github.com/spf13/cobra

Keep in mind that the current scw API is based on docker's one which makes the distinction between global and command-specific flags

I suggest you vote with the rest of the team to choose a norm

PR #509, opened 2018-05-28
Remove VC product line

sieben:
@kdeldycke Could you review the change I brought to the pricing module? I don't know how to get the right data for the START1-XS


kdeldycke:
Ah yes, you're right. `VC1*` instances are no longer available to public consumptions, as per: https://blog.online.net/2018/05/03/introducing-scaleway-nextgen-nvme-cloud-servers-with-hot-snapshots/

LGTM, feel free to merge! :)

PR #520, opened 2018-08-24
WIP: S3 integration

SegFault42:
WIP add s3 command

Integrate minio package to scw

TODO:

- Merge .scwrc and ~/.scw/config.json
- Naming convention (mc to scw)


pkg/config/config.go
```diff
@@ -87,7 +87,7 @@ func GetConfigFilePath() (string, error) {
 	if err != nil {
 		return "", err
 	}
-	return filepath.Join(path, ".scwrc"), nil
+	return filepath.Join(path, ".scw/config.json"), nil
```
QuentinPerez: Will break the legacy ?

QuentinPerez: It works when `~/.scw` doesn't exist ?

justinclift: Should probably also split the string into `.scw` and `config.json` too, instead of embedding the `/` in it.  [`filepath.Join()`](https://golang.org/pkg/path/filepath/#Join) will add the OS appropriate separator.

reversTeam:
See : https://github.com/scaleway/scaleway-cli/pull/529

PR #521, opened 2018-08-27
Add --config option for arbitrary config file

fflorens:
Simple commit to add the ability to use an arbitrary config file instead of `~/.scwrc`

### DIFF
--- a/pkg/cli/main.go
+++ b/pkg/cli/main.go
@@ -48,11 +48,7 @@ func Start(rawArgs []string, streams *commands.Streams) (int, error) {
 	}
 	flag.CommandLine.Parse(rawArgs)
 
-	if *flConfig != "" {
-		os.Setenv("SCW_CONFIG_PATH", *flConfig)
-	}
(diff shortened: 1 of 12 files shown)


sieben:
LGTM

brmzkw:
It would be great to add an unittest

pkg/cli/main.go
```diff
@@ -47,6 +48,10 @@ func Start(rawArgs []string, streams *commands.Streams) (int, error) {
 	}
 	flag.CommandLine.Parse(rawArgs)
 
+	if *flConfig != "" {
+		os.Setenv("SCW_CONFIG_PATH", *flConfig)
```
brmzkw: I think you should not set the environment variable here, but instead return a default if `SCW_CONFIG_PATH` is not set below in `pkg/config/config.go`.

fflorens:
Rebased against master

brmzkw:
thanks! <3

PR #536, opened 2019-01-28
Fix/fuzzysearch path

akerl:
Howdy!

I ran into an issue with go modules, and traced it back to the import here. I'm seeing this error when I use packer, which uses scaleways:

```
go: github.com/renstrom/fuzzysearch@v1.0.2: parsing go.mod: unexpected module path "github.com/lithammer/fuzzysearch"
```

It looks like it's happening because the fuzzysearch repo was migrated to a new owner since it was added here. GitHub helpfully redirects the URL, and most golang tools are OK with that, but the module system compares the import path with the stated names inside the repo and gets unhappy.


reversTeam:
github.com/lithammer/fuzzysearch/fuzzy 404

asfaltboy:
@akerl could you kindly provide steps to reproduce your issue? I just checked out this repository and can build the package as [per instructions in README][1] (using `make`).

Note that github is smart enough to redirect a moved repository:

```
curl -I https://github.com/renstrom/fuzzysearch                                            Sun Oct 13 09:01:22 2019
HTTP/1.1 301 Moved Permanently
Date: Sun, 13 Oct 2019 07:01:30 GMT
Content-Type: text/html; charset=utf-8
Server: GitHub.com
Status: 301 Moved Permanently
Vary: X-PJAX
Location: https://github.com/lithammer/fuzzysearch
...
```

Given this, doing `go get github.com/renstrom/fuzzysearch` actually still works for me, even right now...

[1]: https://github.com/scaleway/scaleway-cli#development

akerl:
The issue occurs when using go Modules, when this repo is a dependency. While GitHub redirects transferred repos, go modules expect the dependency’s internal naming (in go.mod) to match the path used in the go.mod that imports it. This breaks for this repo, because the name in its go.mod still references the old repo.

akerl:
@jerome-quere , I’m not sure why this is closed?

It looks like the issue is still there.

PR #546, opened 2019-03-28
fix: create servers volumes algo

kindermoumoute:
The size of the default volume will now be resized using the following
rules:
1- the user define a custom root size
2- (default) use the largest possible size ==>
`min(categoryMaxSize,volumeMaxSize)`
3- the user specify additional volumes ==> `min(50G,volumeMaxSize)`


frivoire:
LGTM

PR #556, opened 2019-07-01
Small typo in "scw stop" documentation

alexAubin:
I believe `scw wait server` was intended instead of `stop wait server` :wink: 

(Hopefully this is the right way to do a pull request, total newbie in this tool/project)


kindermoumoute:
Good catch @alexAubin, this is a proper way to contribute! 👍

alexAubin:
>This example is also in the readme, can you make the change here too?

Ah indeed ! Here it goes :+1:

PR #562, opened 2019-08-23
add support for exact match when looking for an image

jerome-quere:


QuentinBrosse:
@jerome-quere can you please push an empty commit to relaunch checks :)

PR #584, opened 2019-11-06
chore: allow building in module mode with Go 1.13.3

bcmills:
CC @bradfitz


bcmills:
As noted in the commit message for https://github.com/scaleway/scaleway-cli/pull/584/commits/30eb7560059301e445fe9e8f9dbb7b6840d028ef:
> All packages build at this commit, but not all tests pass.
However, the tests did not all pass before this commit either;
it's not obvious to me whether any regressions are present.

The problem in this case is an existing defect in the `scaleway-cli` Travis configuration: https://github.com/scaleway/scaleway-cli/blob/ef6e81b3a34271b149d1a105801860735f18276f/.travis.yml#L33

`github.com/golang/go/src/cmd/vet` is not the correct import path for that binary. The correct import path, [according to the `go.mod` file for the module containing that binary](https://github.com/golang/go/blob/99957b6930c76b683dbca1ff4bcdd56e59b1e035/src/cmd/go.mod#L1), is `cmd/vet`.

If you want a version of the `vet` binary that is newer than what shipped with the Go toolchain, the appropriate import path for that is `golang.org/x/tools/cmd/vet`.

QuentinBrosse:
Hi @bcmills,
Thanks for your answer. Sorry I didn't notice your commit message.

Here is a fix of the CI pipeline: https://github.com/scaleway/scaleway-cli/pull/585:
> Blocked by moul/anonuuid#30 who upgrades github.com/urfave/cli to v1.22.1 in order to support Go 1.11.

PR #586, opened 2019-12-17
test: add circle ci

jerome-quere:


.circleci/config.yml
```diff
@@ -0,0 +1,24 @@
...
+      - run:
+          name: Execute main binary
+          command: go run cmd/scw/main.go -h
+
+
+workflows:
+  version: 2
+  test:
+    jobs:
+      - test
```
kindermoumoute: nit: EOL

.circleci/config.yml
```diff
@@ -0,0 +1,24 @@
+# Golang CircleCI 2.0 configuration file
+#
+# Check https://circleci.com/docs/2.0/language-go/ for more details
+version: 2
+jobs:
+  test:
+    docker:
+      - image: circleci/golang:1.13
```
QuentinBrosse: Add go1.12?

jerome-quere: We don't need as we are building a binary. It's not a library

QuentinBrosse: Nevertheless useful for contributors. Also, some package manager need a client side compilation.

PR #591, opened 2019-12-17
feat: update generated namespaces

jerome-quere:


internal/namespaces/instance/v1/instance_cli.go
```diff
@@ -470,25 +470,31 @@ func instanceServerUpdate() *core.Command {
...
+			},
+			{
+				Name:       "volumes.{key}.value.id",
 				Short:      `Display the volumes unique ID`,
 				Required:   false,
 				EnumValues: []string{},
 			},
 			{
-				Name:       "volumes.{key}.name",
+				Name:       "volumes.{key}.value.name",
```
kindermoumoute: `.value` should not appear here

PR #593, opened 2019-12-18
chore: return typed error in arg package

loicbourgois:
Internal ticket: APIGW-1671

### DIFF
--- a/internal/args/errors.go
+++ b/internal/args/errors.go
@@ -3,6 +3,8 @@ package args
 import (
 	"fmt"
 	"reflect"
+	"strconv"
+	"strings"
 )
 
 //
@@ -52,7 +54,7 @@ type InvalidArgumentError struct {
 }
 
 func (e *InvalidArgumentError) Error() string {
-	return fmt.Sprintf("invalid argument '%s': must only contain lowercase letter, number or dash", e.ArgumentName)
+	return fmt.Sprintf("invalid argument '%s': must only contain lowercase letters, numbers or dashes", e.ArgumentName)
 }
 
 type UnknowArgumentError struct {
@@ -110,20 +112,17 @@ func (e *InvalidIndexError) Error() string {
 }
 
 type MissingIndicesInArrayError struct {
-	Index          uint64
-	MissingIndices string
+	IndexToInsert int
+	CurrentLength int
 }
 
 func (e *MissingIndicesInArrayError) Error() string {
-	return fmt.Sprintf("missing indices in the array: trying to set array at index %d before indices %s", e.Index, e.MissingIndices)
-}
-
-type MissingIndexInArrayError struct {
-	Index uint64
-}
-
-func (e *MissingIndexInArrayError) Error() string {
-	return fmt.Sprintf("missing index in the array: trying to set array at index %d before index %d", e.Index, e.Index-1)
+	switch {
+	case e.IndexToInsert-e.CurrentLength == 1:
+		return fmt.Sprintf("missing index %d: all indices prior to %d must be set as well", e.CurrentLength, e.IndexToInsert)
+	default:
+		return fmt.Sprintf("missing indices %s: all indices prior to %d must be set as well", missingIndices(int(e.IndexToInsert), e.CurrentLength), e.IndexToInsert)
+	}
 }
 
 type NoSubKeyForMapError struct {
@@ -164,4 +163,20 @@ type CannotUnmarshalError struct {
 
 func (e *CannotUnmarshalError) Error() string {
 	return fmt.Sprintf("%T is not unmarshalable", e.Interface)
+}
+
+// missingIndices returns a string of all the missing indices between index and length.
+// e.g.: missingIndices(index=5, length=0) should return "0,1,2,3"
+// e.g.: missingIndices(index=5, length=2) should return "2,3"
+// e.g.: missingIndices(index=99999, length=0) should return "0,1,2,3,4,5,6,7,8,9,..."
+func missingIndices(index, length int) string {
+	s := []string(nil)
+	for i := length; i < index; i++ {
+		if i-length == 10 {
+			s = append(s, "...")
+			break
+		}
+		s = append(s, strconv.Itoa(i))
+	}
+	return strings.Join(s, ",")
 }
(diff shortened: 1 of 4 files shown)


internal/args/errors.go
```diff
@@ -0,0 +1,167 @@
...
+	return fmt.Sprintf("invalid index: '%s' is not a positive integer", e.Index)
+}
+
+type MissingIndicesInArrayError struct {
+	Index          uint64
+	MissingIndices string
+}
+
+func (e *MissingIndicesInArrayError) Error() string {
+	return fmt.Sprintf("missing indices in the array: trying to set array at index %d before indices %s", e.Index, e.MissingIndices)
```
kindermoumoute: missing indices should be computed in this func

internal/args/errors.go
```diff
@@ -0,0 +1,167 @@
...
+	Index string
+}
+
+func (e *InvalidIndexError) Error() string {
+	return fmt.Sprintf("invalid index: '%s' is not a positive integer", e.Index)
+}
+
+type MissingIndicesInArrayError struct {
+	Index          uint64
+	MissingIndices string
```
kindermoumoute: Should have Length field instead of MissingIndices

internal/args/errors.go
```diff
@@ -0,0 +1,167 @@
...
+
+func (e *MissingIndicesInArrayError) Error() string {
+	return fmt.Sprintf("missing indices in the array: trying to set array at index %d before indices %s", e.Index, e.MissingIndices)
+}
+
+type MissingIndexInArrayError struct {
+	Index uint64
+}
+
+func (e *MissingIndexInArrayError) Error() string {
```
kindermoumoute: This error is a duplicate of `MissingIndicesInArrayError`

internal/args/errors.go
```diff
@@ -0,0 +1,167 @@
...
+	return fmt.Sprintf("missing index in the array: trying to set array at index %d before index %d", e.Index, e.Index-1)
```
kindermoumoute: error message should compute the plural form

internal/args/errors.go
```diff
@@ -0,0 +1,167 @@
...
+
+//
+// Marshal Errors
+//
+
+type DataMustBeAmArshalableValueError struct {
+}
+
+func (e *DataMustBeAmArshalableValueError) Error() string {
+	return fmt.Sprintf("data must be a marshable value (a scalar type or a Marshaler)")
```
kindermoumoute: Should not use struct for static message
```
ErrDataMustBeAmArshalableValue = fmt.Errorf("data must be a marshable value (a scalar type or a Marshaler)")
```

internal/args/errors.go
```diff
@@ -0,0 +1,167 @@
...
+
+//
+// Marshal & Unmarshal Errors
+//
+
+type DataMustBeAPointerError struct {
+}
+
+func (e *DataMustBeAPointerError) Error() string {
+	return fmt.Sprintf("data must be a pointer to a struct")
```
kindermoumoute: Should not use struct for static message
```
ErrDataMustBeAPointer = fmt.Errorf("data must be a pointer to a struct")
```

internal/args/errors.go
```diff
@@ -0,0 +1,167 @@
...
+
+func (e *DuplicateArgumentError) Error() string {
+	return fmt.Sprintf("duplicate argument '%s'", e.ArgumentName)
+}
+
+type DataIsNilError struct {
+}
+
+func (e *DataIsNilError) Error() string {
+	return fmt.Sprintf("data must be not be nil")
```
kindermoumoute: Should not use struct for static message
```
ErrDataIsNil = fmt.Errorf("data must be not be nil")
```

internal/args/errors.go
```diff
@@ -0,0 +1,167 @@
...
+}
+
+type DataIsNotAPointerError struct {
+}
+
+func (e *DataIsNotAPointerError) Error() string {
+	return fmt.Sprintf("data must be a pointer")
```
kindermoumoute: Should not use struct for static message
```
ErrDataIsNotAPointer = fmt.Errorf("data must be a pointer")
```

internal/args/errors.go
```diff
@@ -0,0 +1,167 @@
...
+
+func (e *CannotSetNestedFieldError) Error() string {
+	return fmt.Sprintf("cannot set nested field %s for unmarshalable type %T", e.ArgumentName, e.Interface)
+}
+
+type MissingIndexOnArrayError struct {
+}
+
+func (e *MissingIndexOnArrayError) Error() string {
+	return fmt.Sprintf("missing index on the array")
```
kindermoumoute: Should not use struct for static message
```
ErrMissingIndexOnArray = fmt.Errorf("missing index on the array")
```

internal/args/unmarshal_test.go
```diff
@@ -99,15 +99,15 @@ func TestUnmarshalStruct(t *testing.T) {
 			"unknown-field=12",
 		},
 		expected: &Basic{},
-		error:    "unknown argument 'unknown-field'",
+		error:    "unknown argument with name 'unknown-field'",
```
kindermoumoute: I am not entirely convinced `with name` provides any information since `unknown-field` is called an argument

internal/args/errors.go
loicbourgois: I thought we agreed to use struct for all errors
> All errors, whether they are useful for the end user or not, should follow the format presented bellow.
> pro: we don't need to think about how to group errors, and can concentrate on more important problems
> pro: we can use a single switch to reformat some errors if needed

internal/args/unmarshal.go
```diff
@@ -211,10 +211,8 @@ func set(dest reflect.Value, argNameWords []string, value string) error {
 		// Make sure array is big enough to access the correct index.
 		diff := int(index) - dest.Len()
 		switch {
-		case diff > 2:
-			return &MissingIndicesInArrayError{Index: index, MissingIndices: missingIndices(int(index), dest.Len())}
-		case diff == 1:
-			return &MissingIndexInArrayError{index}
+		case diff >= 1:
```
kindermoumoute: nit: `diff >0`

PR #594, opened 2019-12-18
fix: remove warnings for fields not ignored

loicbourgois:
```
> scw instance ip create -h
DEBUG: 2019/12/18 15:21:10 building instance.server.update: '[volumes {key} id]' has no usage and is not ignored
DEBUG: 2019/12/18 15:21:10 building instance.image.create: '[extra-volumes {key} id]' has no usage and is not ignored
Reserve an IP.

USAGE:
  scw [global-flags] instance ip create [flags] [args]

ARGS:
  [server]   UUID of the server you want to attach the IP to

FLAGS:
  -h, --help   help for create

GLOBAL FLAGS:
      --access-key string        Scaleway access key
  -D, --debug                    Enable debug mode
      --organization-id string   Scaleway organization ID
  -o, --output PrinterType       Output format: json or human
  -p, --profile string           The config profile to use
  -r, --region string            The Scaleway region
      --secret-key string        Scaleway secret key
  -z, --zone string              The Scaleway zone

### DIFF
--- a/internal/namespaces/instance/v1/instance_cli.go
+++ b/internal/namespaces/instance/v1/instance_cli.go
@@ -20,14 +20,12 @@ func GetGeneratedCommands() *core.Commands {
 		instancePlacementGroup(),
 		instancePlacementGroupServer(),
 		instanceSecurityGroup(),
-		instanceSecurityGroupRule(),
 		instanceServer(),
 		instanceServerType(),
 		instanceSnapshot(),
 		instanceVolume(),
 		instanceServerTypeList(),
 		instanceServerList(),
-		instanceServerDelete(),
 		instanceServerGet(),
 		instanceServerUpdate(),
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: internal/namespaces/instance/v1/instance_cli.go [hunk 1]
```diff
@@ -438,25 +439,31 @@ func instanceServerUpdate() *core.Command {
...
 			},
 			{
-				Name:       "volumes.{key}.id",
+				Name:       "volumes.{key}.key",
+				Short:      ``,
+				Required:   false,
+				EnumValues: []string{},
+			},
+			{
+				Name:       "volumes.{key}.value.id",
```
kindermoumoute: same as https://github.com/scaleway/scaleway-cli/pull/591, generation should be fixed before

PR #597, opened 2019-12-19
chore: make panic on e2e tests more explicit

loicbourgois:


internal/core/testing.go
```diff
@@ -103,16 +103,18 @@ func Test(config *TestConfig) func(t *testing.T) {
 		client := getTestClient(t, config.UseE2EClient)
 
 		executeCmd := func(cmd string) {
+			stdoutBuffer := &bytes.Buffer{}
```
jerome-quere: you could use stdout and stderr as long as you call reset before.

loicbourgois: This would make it confusing. 

I think it's better to keep the buffer as close to the execution as possible. 

We have 3 call to Bootstrap(), we should have 3 pairs of buffers: pre, command, post.

Reseting the same buffers will be confusing.

PR #599, opened 2019-12-19
feat(core): add support for raw example

jerome-quere:


internal/interactive/print.go
```diff
@@ -55,7 +55,10 @@ func Indent(str string, indent int) string {
 	lines := strings.Split(str, "\n")
 	result := ""
 	for _, line := range lines {
-		result += padding + line + "\n"
+		if line != "" {
```
kindermoumoute: should it avoid `+ "\n"` when line is empty ?

internal/core/cobra_usage_builder.go
```diff
@@ -124,28 +125,40 @@ func buildExamples(cmd *Command) string {
 	for _, cmdExample := range cmd.Examples {
 		// Build title.
 		title := fmt.Sprintf("  %s", cmdExample.Short)
+		commandLine := ""
+		switch {
```
loicbourgois: maybe `case cmdExample.Raw != "" && case cmdExample.Raw != "" ` should trigger a warning

Nothing prevents a user to set both.

internal/namespaces/test/v1/test_cli.go
```diff
@@ -350,5 +350,13 @@ func testHumanRun() *core.Command {
 			return api.RunHuman(args)
 
 		},
+		Examples: []*core.Example{
+			{
+				Short: "Create a human and make it run",
+				Raw: `scw test human create
+scw test human run human-id=xxxxx
+`,
```
loicbourgois: shouldn't the generator remove empty lines at the end ?

internal/core/cobra_usage_builder.go
QuentinBrosse: Do you mean: `case cmdExample.Raw != "" && cmdExample.Request != ""` ?

It could be done in the generator validation and here keep the priority for the Raw example here ?

jerome-quere: could be done in the diagnostic tools

internal/interactive/print.go
jerome-quere: When working with a paragraph you may want to include empty line separator
```
  this is a line
<-- no space at the beginning of this line
  this is a line2
```

internal/namespaces/test/v1/test_cli.go
jerome-quere: I created an issue to fix this one on the generator

PR #600, opened 2019-12-19
feat(core): inline marshalling for scalar slices

QuentinBrosse:
* Implement `marshalInlineSlice`
* Remove `serverSummariesMarshallerFunc` as it is now the default behavior
* Add `Tags` to `humanServerInList` 🎁 

Result:
```
$ scw instance security-group list
ID                                    NAME                    [...]  SERVERS  STATEFUL
00000000-0000-1000-0000-000000000000  Default security group  [...]  3        false
22222222-2222-1222-2222-222222222222  test01                  [...]  0        true
                                                                     ^ a count representation as the ServerSummary is not a scalar type.

$ scw instance server get server-id=64c3ed5c-72f6-495b-94f5-f43c4536f49c
Server:
id                        00000000-0000-1000-0000-000000000000
name                      terraform-v2-doc
organization              11111111-1111-1111-1111-111111111111
allowed-actions.0         backup
tags.0                    tf
tags.1                    test1
tags.2                    test2
                          ^ a {label}.{idx} representation as it is a struct{Slice: []interface{}}
[...]

Server Image:
[...]

$ scw instance server list
ID                                    TAGS              VOLUMES
00000000-0000-1000-0000-000000000000  [tf test1 test2]  1      
                                      ^ an inline slice representation as it's elements are scalar

                                                        ^ use volumeMapMarshallerFunc
```

### DIFF
--- a/internal/human/marshal.go
+++ b/internal/human/marshal.go
@@ -265,7 +265,7 @@ func marshalSlice(slice reflect.Value, opt *MarshalOpt) (string, error) {
 	return formatGrid(grid)
 }
 
-// marshalInlineSlice transform nested scalar slices in an inline string representation
+// marshalInlineSlice transforms nested scalar slices in an inline string representation
 // and other types of slices in a count representation.
 func marshalInlineSlice(slice reflect.Value) (string, error) {
 	itemType := slice.Type().Elem()
--- a/internal/human/marshal.go
+++ b/internal/human/marshal.go
@@ -240,7 +240,7 @@ func marshalSlice(slice reflect.Value, opt *MarshalOpt) (string, error) {
 			fieldValue := getFieldValue(item, fieldSpec.FieldName)
 			str := ""
 			if fieldValue.IsValid() {
-				err := error(nil)
+				var err error
 				switch {
 				// Handle inline slice.
 				case fieldValue.Type().Kind() == reflect.Slice:
--- a/internal/human/marshal.go
+++ b/internal/human/marshal.go
@@ -237,10 +237,24 @@ func marshalSlice(slice reflect.Value, opt *MarshalOpt) (string, error) {
 		item := slice.Index(i)
 		row := []string(nil)
 		for _, fieldSpec := range opt.Fields {
-			value := getFieldValue(item, fieldSpec.FieldName)
+			fieldValue := getFieldValue(item, fieldSpec.FieldName)
 			str := ""
-			if value.IsValid() {
-				str, _ = Marshal(value.Interface(), opt)
+			if fieldValue.IsValid() {
+				err := error(nil)
+				switch {
+				// Handle inline slice.
+				case fieldValue.Type().Kind() == reflect.Slice:
+					str, err = marshalInlineSlice(fieldValue)
+					if err != nil {
+						return "", err
+					}
+
+				default:
+					str, err = Marshal(fieldValue.Interface(), opt)
+					if err != nil {
+						return "", err
+					}
+				}
 			} else {
 				logger.Debugf("invalid getFieldValue(): '%v' might not be exported", fieldSpec.FieldName)
 			}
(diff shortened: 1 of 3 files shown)


internal/human/marshal.go
```diff
@@ -237,10 +237,24 @@ func marshalSlice(slice reflect.Value, opt *MarshalOpt) (string, error) {
 		item := slice.Index(i)
 		row := []string(nil)
 		for _, fieldSpec := range opt.Fields {
-			value := getFieldValue(item, fieldSpec.FieldName)
+			fieldValue := getFieldValue(item, fieldSpec.FieldName)
 			str := ""
-			if value.IsValid() {
-				str, _ = Marshal(value.Interface(), opt)
+			if fieldValue.IsValid() {
+				err := error(nil)
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

internal/human/marshal.go
```diff
@@ -237,10 +237,24 @@ func marshalSlice(slice reflect.Value, opt *MarshalOpt) (string, error) {
 		item := slice.Index(i)
 		row := []string(nil)
 		for _, fieldSpec := range opt.Fields {
-			value := getFieldValue(item, fieldSpec.FieldName)
+			fieldValue := getFieldValue(item, fieldSpec.FieldName)
 			str := ""
-			if value.IsValid() {
-				str, _ = Marshal(value.Interface(), opt)
+			if fieldValue.IsValid() {
+				var err error
```
loicbourgois: why is this necessary ? 
can't we use `:=` notation below ?

internal/human/marshal.go
```diff
@@ -251,6 +265,32 @@ func marshalSlice(slice reflect.Value, opt *MarshalOpt) (string, error) {
...
+		itemType = itemType.Elem()
+	}
+
+	switch {
+	// If marshaler func is available.
+	// As we cannot set MarshalOpt of a nested slice opt will always be nil here.
+	case marshalerFuncs[itemType] != nil:
+		return marshalerFuncs[itemType](slice.Interface(), nil)
+
+	// If it's a scalar value we marshall it inline.
```
loicbourgois: Isn't it `// If the slice contains scalar values ` ?

internal/human/marshal.go
QuentinBrosse: Nop, `str` is already defined. https://play.golang.com/p/NwyLo8zmcMm

loicbourgois:
```
$ scw instance server get server-id=64c3ed5c-72f6-495b-94f5-f43c4536f49c
Server:
id                        00000000-0000-1000-0000-000000000000
name                      terraform-v2-doc
organization              11111111-1111-1111-1111-111111111111
allowed-actions.0         backup
tags.0                    tf
tags.1                    test1
tags.2                    test2
                          ^ a {label}.{idx} representation as it is a struct{Slice: []interface{}}
[...]
```

For `tags`, shouldn't it be on a single line ? 
```
tags                      tf, test1, test2

File: internal/human/marshal.go [hunk 4]
```diff
@@ -251,6 +265,32 @@ func marshalSlice(slice reflect.Value, opt *MarshalOpt) (string, error) {
 	return formatGrid(grid)
 }
 
+// marshalInlineSlice transform nested scalar slices in an inline string representation
```
kindermoumoute: transforms

QuentinBrosse:
@loicbourgois 
The count is useful for lists views, where we must be concise. We are not in a list here so we have space to display long lists. For this reason, IMO, it's better to don't print count for the default behavior on this kind of struct views (eg. `get` commands)

The `tf, test1, test2` could be a great representation too but can't handle long lists.

loicbourgois:
I don't undertsnad why we are not using `tf, test1, test2` representation in `list` command ? 

We usually have more space with a `get` than with a `list`. 

If inline representation is ok for a `list`, it should be ok for a `get` too.

loicbourgois:
@QuentinBrosse also, I don't understand why the tags, which are represented as `[]string` in the sdk are in `[]interface{}` in your example.

see https://github.com/scaleway/scaleway-sdk-go/blob/master/api/instance/v1/instance_sdk.go#L957

QuentinBrosse:
@loicbourgois 
> I don't undertsnad why we are not using tf, test1, test2 representation in list command ?

We are using it: see tags column in `scw instance server list`.

> We usually have more space with a get than with a list.

Exactly, that's why I prefer the `{key}.{index}` representation when we have much vertical and horizontal space. The `{key}.{index}` representation handle slices of structs (and more) contrary to `item1, item2` who can handle only slices of scalar.
Also, if we have a slice of 20 elements, the `{key}.{index}` will take two lines for most of the screens.

> I don't understand why the tags, which are represented as []string in the sdk are in []interface{} in your example.

By `[]interface{}` I mean "slice of anything"

PR #601, opened 2019-12-19
fix(instance): fix security group human marshal style

QuentinBrosse:


internal/namespaces/instance/v1/custom_marshal.go
```diff
@@ -325,23 +325,25 @@ func (sg *customSecurityGroupResponse) MarshalHuman() (out string, err error) {
 	if err != nil {
 		return "", err
 	}
+	defaultInboundPolicy = terminal.Style(defaultInboundPolicy, color.Bold)
```
loicbourgois: If the whole title is bold, we should try to use `color.Bold` only once per title. 

Wouldn't something like that work ? 
```
inboundRulesTitle := terminal.Style("Inbound Rules (default policy " + defaultInboundPolicy + "):", color.Bold)
```

and at line 340 

```
inboundRulesView := inboundRulesTitle + "\n" + inboundRulesContent

Comment (QuentinBrosse) [reply internal/namespaces/instance/v1/custom_marshal.go]: Nop, the `defaultInboundPolicy` (in your fist code section) will escape at the end and the `):` will not be bold. :/

Comment (loicbourgois) [reply internal/namespaces/instance/v1/custom_marshal.go]: And what about ?
```
inboundRulesTitle := "Inbound Rules (default policy " + defaultInboundPolicy + "):"
inboundRulesView := terminal.Style(inboundRulesTitle, color.Bold) + "\n" + inboundRulesContent

loicbourgois: ho, ok, defaultInboundPolicy is already colored

Sorry, i missed that

loicbourgois: Can you add a comment please ? 

It's not clear `defaultInboundPolicy` is already colored, wich leads to the escaping problem.

internal/namespaces/instance/v1/custom_marshal.go
```diff
@@ -325,23 +325,25 @@ func (sg *customSecurityGroupResponse) MarshalHuman() (out string, err error) {
...
 		return "", err
 	}
+	defaultOutboundPolicy = terminal.Style(defaultOutboundPolicy, color.Bold)
 
 	inboundRulesContent, err := human.Marshal(inboundRules, nil)
 	if err != nil {
 		return "", err
 	}
-	inboundRulesView := terminal.Style(fmt.Sprintf("Inbound Rules (default policy %s):\n", defaultInboundPolicy), color.Bold) + inboundRulesContent
+	inboundRulesView := terminal.Style("Inbound Rules (default policy ", color.Bold) + defaultInboundPolicy + terminal.Style("):\n", color.Bold) + inboundRulesContent
```
loicbourgois: and here, it's not clear `defaultInboundPolicy` is already bold

Maybe remove `defaultInboundPolicy = terminal.Style(defaultInboundPolicy, color.Bold)` at l328, and here 

```
// comment explaning teh escaping problem
inboundRulesView := terminal.Style("Inbound Rules (default policy ", color.Bold) + 
    terminal.Style(defaultInboundPolicy,  color.Bold) + 
    terminal.Style("):", color.Bold) + 
    "\n" +  inboundRulesContent

PR #606, opened 2019-12-23
feat(core): add custom marshaler for ResourceExpired

remyleone:


internal/core/error.go
```diff
@@ -162,3 +163,14 @@ func sdkOutOfStockErrorHumanMarshallerFunc() human.MarshalerFunc {
...
 	}
 }
+
+func sdkResourceExpiredHumanMarshallFunc() human.MarshalerFunc {
+	return func(i interface{}, opt *human.MarshalOpt) (string, error) {
+		resourceExpiredError := i.(*sdk.ResourceExpiredError)
+
+		return human.Marshal(&CliError{
+			Err:  fmt.Errorf("resource %s with ID %s expired since %s", resourceExpiredError.Resource, resourceExpiredError.ResourceID, resourceExpiredError.ExpiredSince.String()),
+			Hint: "Try to generate a new token.",
```
QuentinBrosse: If the resource is not a token, it could be a strange hint..

internal/core/error.go
```diff
@@ -162,3 +163,22 @@ func sdkOutOfStockErrorHumanMarshallerFunc() human.MarshalerFunc {
...
 }
+
+func sdkResourceExpiredHumanMarshallFunc() human.MarshalerFunc {
+	return func(i interface{}, opt *human.MarshalOpt) (string, error) {
+		resourceExpiredError := i.(*sdk.ResourceExpiredError)
+
+		var hint string
+		switch resourceName := resourceExpiredError.Resource; resourceName {
+		case "account_token":
+			hint = "Try to generate a new token here [URL credential de la console]"
```
kindermoumoute: `[URL credential de la console]` ==> `https://console.scaleway.com/account/credentials`

PR #616, opened 2020-01-07
fix(args): rework slices

kindermoumoute:
Today it is not very explicit for a user because he can use 3 methods to fill slice:

```
slice.0=a -slice.1=b
slice=a -slice=b
slice=a,b (In custom commands like create server)
```

This PRs does:

- [x] Remove comma notation (in create server)
- [x] Remove support for slice without index (throw error)
- [x] Remove support for auto-filling slice on missing index and return  an error instead
- [x] Fix a bug that recreated a new slice for each index, which destroyed all previous elements in the slice

### DIFF
--- a/internal/args/args.go
+++ b/internal/args/args.go
@@ -7,7 +7,7 @@ import (
 )
 
 // validArgNameRegex regex to check that args words are lower-case or digit starting and ending with a letter.
-var validArgNameRegex = regexp.MustCompile("^([a-z][a-z0-9-]*)(\\.[a-z0-9-]*)*$")
+var validArgNameRegex = regexp.MustCompile(`^([a-z][a-z0-9-]*)(\.[a-z0-9-]*)*$`)
 
 // RawArgs allows to retrieve a simple []string using UnmarshalStruct()
 type RawArgs []string
--- a/internal/args/args.go
+++ b/internal/args/args.go
@@ -6,8 +6,8 @@ import (
 	"strings"
 )
 
-// validKeyRegex regex to check that args key are lower-case or digit starting and ending with a letter.
-var validKeyRegex = regexp.MustCompile("^([a-z][a-z0-9-]*)(\\.[a-z0-9-]*)*$")
+// validArgNameRegex regex to check that args words are lower-case or digit starting and ending with a letter.
+var validArgNameRegex = regexp.MustCompile("^([a-z][a-z0-9-]*)(\\.[a-z0-9-]*)*$")
 
 // RawArgs allows to retrieve a simple []string using UnmarshalStruct()
 type RawArgs []string
--- a/internal/args/unmarshal.go
+++ b/internal/args/unmarshal.go
@@ -2,13 +2,12 @@ package args
 
 // unmarshal.go helps with the conversion of
 // CLI arguments represented as strings
-// into CLI arguments represented as go data.
+// into CLI arguments represented as Go data.
 
 import (
 	"fmt"
 	"io"
 	"reflect"
-	"regexp"
 	"strconv"
 	"strings"
 
@@ -41,10 +40,10 @@ var unmarshalFuncs = map[reflect.Type]UnmarshalFunc{
 	},
 }
 
-// UnmarshalStruct parse args like ["arg1=1", "arg2=2"] to a go structure using reflection.
+// UnmarshalStruct parse args like ["arg1=1", "arg2=2"] to a Go structure using reflection.
 //
 // args: slice of args passed through the command line
-// data: go structure to fill
+// data: Go structure to fill
 func UnmarshalStruct(args []string, data interface{}) error {
 
 	// First check if we want to retrieve a simple []string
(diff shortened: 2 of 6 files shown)


internal/args/args.go
```diff
@@ -6,8 +6,8 @@ import (
 	"strings"
 )
 
-// validKeyRegex regex to check that args key are lower-case or digit starting and ending with a letter.
-var validKeyRegex = regexp.MustCompile("^([a-z][a-z0-9-]*)(\\.[a-z0-9-]*)*$")
+// validArgNameRegex regex to check that args words are lower-case or digit starting and ending with a letter.
+var validArgNameRegex = regexp.MustCompile("^([a-z][a-z0-9-]*)(\\.[a-z0-9-]*)*$")
```
golangcibot: S1007: should use raw string (`...`) with regexp.MustCompile to avoid having to escape twice (from `gosimple`)

internal/args/unmarshal.go
```diff
@@ -63,36 +62,32 @@ func UnmarshalStruct(args []string, data interface{}) error {
 
 	// Map arg names to their values.
 	// ["arg1=1", "arg2=2", "arg3"] => {"arg1": "1", "arg2": "2", "arg3":"" }
```
jerome-quere: Comment do not reflect what happend

internal/args/unmarshal.go
```diff
@@ -101,31 +96,31 @@ func UnmarshalStruct(args []string, data interface{}) error {
 	return nil
 }
 
-func findFieldType(t reflect.Type, keys []string) (reflect.Type, error) {
+func fieldExist(t reflect.Type, argNameWords []string) bool {
 
 	switch {
-	case len(keys) == 0:
-		return t, nil
+	case len(argNameWords) == 0:
```
jerome-quere: In this case we should also check that we know how to unmarshal a type T.

This wil handle the case where argname should be `a.b.c.d=12` but we only type `a.b=12`

internal/args/unmarshal.go
```diff
@@ -190,70 +185,95 @@ func set(dest reflect.Value, keys []string, value string) error {
...
-		newValue := reflect.New(dest.Type().Elem())
-		err := set(newValue.Elem(), keys, value)
-		dest.Set(reflect.Append(dest, newValue.Elem()))
-		return err
+		// Make sure array is big enough to access the correct index.
+		diff := int(index) - dest.Len()
+		switch {
+		case diff > 2:
+			return fmt.Errorf("missing indices in the array: trying to set array at index %d before indices %v", index, missingIndices(int(index), dest.Len()))
+		case diff == 1:
```
jerome-quere: could be handle in a single case with the one before

internal/args/unmarshal.go
```diff
@@ -190,70 +185,95 @@ func set(dest reflect.Value, keys []string, value string) error {
...
-		// We create a new value call set and append it to the slice
-		newValue := reflect.New(dest.Type().Elem())
-		err := set(newValue.Elem(), keys, value)
-		dest.Set(reflect.Append(dest, newValue.Elem()))
-		return err
+		// Make sure array is big enough to access the correct index.
+		diff := int(index) - dest.Len()
+		switch {
+		case diff > 2:
+			return fmt.Errorf("missing indices in the array: trying to set array at index %d before indices %v", index, missingIndices(int(index), dest.Len()))
```
jerome-quere: I prefer `%s` for string

internal/args/unmarshal.go
```diff
@@ -190,70 +185,95 @@ func set(dest reflect.Value, keys []string, value string) error {
...
+}
+
+// missingIndices returns a string of all the missing indices between index and length.
+// e.g.: missingIndices(index=5, length=0) should return "3,2,1,0"
+// e.g.: missingIndices(index=5, length=2) should return "3,2"
+// e.g.: missingIndices(index=99999, length=0) should return "99998,99997,99996,99995,99994,99993,99992,99991,99990,..."
+func missingIndices(index, length int) string {
+	s := ""
+	for i := index - 1; i >= length; i-- {
+		if s != "" {
```
jerome-quere: why not computing a slice and call `strings.Join`

internal/args/unmarshal.go
```diff
@@ -315,9 +335,13 @@ func isUnmarshalableValue(dest reflect.Value) bool {
 
 	_, isUnmarshaller := interface_.(Unmarshaller)
 	_, hasUnmarshalFunc := unmarshalFuncs[dest.Type()]
-	_, isScalar := scalarKinds[dest.Kind()]
 
-	return isUnmarshaller || hasUnmarshalFunc || isScalar
+	return isUnmarshaller || hasUnmarshalFunc || isScalar(dest)
```
jerome-quere: why create a function for `isScalar` ans not the 2 others ?

internal/args/unmarshal_test.go
```diff
@@ -110,30 +110,77 @@ func TestUnmarshalStruct(t *testing.T) {
...
+		expected: &Slice{},
+		error:    "missing index in the array: trying to set array at index 1 before index 0",
+	}))
+
+	t.Run("missing-slice-indices", run(TestCase{
+		args: []string{
+			"strings.5=2",
+		},
+		expected: &Slice{},
+		error:    "missing indices in the array: trying to set array at index 5 before indices 4,3,2,1,0",
```
jerome-quere: why printing index in reverse 🤔

internal/args/unmarshal.go
kindermoumoute: This is already done at the beginning of the `set`

internal/args/unmarshal.go
kindermoumoute: When we'll switch to a typed struct this will make a lot of sense to do this logic in the `Error()` func

internal/args/unmarshal.go
kindermoumoute: It is more code

internal/args/unmarshal.go
kindermoumoute: oops, removing this leftover

PR #638, opened 2020-01-15
feat(core): bad size unit error

QuentinBrosse:
Internal Ticket: APIGW-1674

### DIFF
--- a/internal/args/errors.go
+++ b/internal/args/errors.go
@@ -21,11 +21,11 @@ func (e *DataMustBeAPointerError) Error() string {
 // Marshal Errors
 //
 
-type DataMustBeAmArshalableValueError struct {
+type DataMustBeAMarshalableValueError struct {
 }
 
-func (e *DataMustBeAmArshalableValueError) Error() string {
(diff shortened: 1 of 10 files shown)


internal/args/errors.go
```diff
@@ -180,3 +180,11 @@ func missingIndices(index, length int) string {
 	}
 	return strings.Join(s, ",")
 }
+
+type InvalidSizeUnitError struct {
+	Value string
```
kindermoumoute: looks like this value is not used

internal/args/unmarshal.go
```diff
@@ -23,6 +23,12 @@ type UnmarshalFunc func(value string, dest interface{}) error
 
 var unmarshalFuncs = map[reflect.Type]UnmarshalFunc{
 	reflect.TypeOf((*scw.Size)(nil)).Elem(): func(value string, dest interface{}) error {
+		// Only support G, GB for now (case insensitive).
+		value = strings.ToLower(value)
+		if !strings.HasSuffix(value, "g") && !strings.HasSuffix(value, "gb") {
+			return &InvalidSizeUnitError{}
```
kindermoumoute: field value should be filled

PR #639, opened 2020-01-15
refactor(instance): custom commands, marshalers and overrides

QuentinBrosse:
Internal Ticket: APIGW-1725


internal/core/command.go
```diff
@@ -70,6 +70,16 @@ type WaitFunc func(ctx context.Context, argsI, respI interface{}) error
 
 const indexCommandSeparator = "."
 
+// OverrideRun overrides the CommandRunner function via a builder function.
+func (c *Command) OverrideRun(runBuilder func(command *Command) CommandRunner) {
```
jerome-quere: May be we should have a singel Override method 
```
cmds.MustFind("instance", "placement-group", "get").Override(placementGroupGetOverride)
```

loicbourgois: Why do we even need `OverrideRun()` or `Override()` ? 

Can't we just do `myCommand.run = overrideMyCommandRun(myCommand)` ?

Or `myCommand = overrideMyCommand(myCommand)`

A general `OverrideRun()` could be useful if we're doing more than justre reassigning `c.Run`, like logging the fact that something has been overriden. But as is, I don't understand why we need a whole function for `c.Run = runBuilder(c)`.

internal/namespaces/instance/v1/custom_server.go
```diff
@@ -0,0 +1,358 @@
...
+}
+
+// serverLocationMarshalerFunc marshals a instance.ServerLocation.
+func serverLocationMarshalerFunc(i interface{}, opt *human.MarshalOpt) (string, error) {
+	location := i.(instance.ServerLocation)
+	zone, err := scw.ParseZone(location.ZoneID)
+	if err != nil {
+		return "", err
+	}
+	zoneStr := fmt.Sprintf("%s", zone)
```
golangcibot: S1025: the argument's underlying type is a string, should use a simple conversion instead of fmt.Sprintf (from `gosimple`)

PR #642, opened 2020-01-16
fix: better handling of Ctrl-C and Ctrl-D in Readline

loicbourgois:


internal/interactive/prompt.go
```diff
@@ -153,7 +153,7 @@ func Readline(config *ReadlineConfig) (string, error) {
 			case "Interrupt":
 				return "", &InterruptError{}
```
jerome-quere: I think it would be better to return a CliError with an exit code here

internal/core/bootstrap.go
```diff
@@ -70,6 +71,12 @@ func Bootstrap(config *BootstrapConfig) (exitCode int) {
 
 	rootCmd.SetArgs(config.Args[1:])
 	err = rootCmd.Execute()
+
+	switch err.(type) {
```
jerome-quere: we should add support for an ExitCode field inside CLI error. and return 1 by default

loicbourgois: where should we transform `interactive.InterruptError` into a `CliError` ?

internal/interactive/prompt.go
loicbourgois: are we sure we want to return a `CliError` from `interactive` ?

internal/core/bootstrap.go
```diff
@@ -70,6 +71,13 @@ func Bootstrap(config *BootstrapConfig) (exitCode int) {
...
+	case *interactive.InterruptError:
+		return 130
+	}
+	fmt.Println("e", err)
```
jerome-quere: shoud be removed

internal/core/bootstrap_test.go
```diff
@@ -0,0 +1,35 @@
...
+import (
+	"context"
+	"reflect"
+	"testing"
+
+	"github.com/scaleway/scaleway-cli/internal/args"
+	"github.com/scaleway/scaleway-cli/internal/interactive"
+)
+
+func getCommands() *Commands {
```
jerome-quere: this "pollute" core test package let's write:
```
 func TestInterruptError(t *testing.T) {
  commands := NewCommands(&Command{...})
  t.Run(....)
}
```

PR #648, opened 2020-01-20
fix(autocomplete): only save args into completedArgs

loicbourgois:


internal/core/autocomplete.go
```diff
@@ -234,7 +234,8 @@ func AutoComplete(ctx context.Context, leftWords []string, wordToComplete string
 
 	// For each left word that is not a flag nor an argument, we try to go deeper in the autocomplete tree and store the current node in `node`.
 	node := commandTreeRoot
-	for _, word := range leftWords {
+	nodeIndexInWords := 0
```
jerome-quere: // lastCmdWordIndex is the leftmost word index that contain either a namespace a ressource or command.
lastCmdWordIndex :=

PR #649, opened 2020-01-20
feat(args): handle anonymous fields in UnmarshalStruct

jerome-quere:


internal/args/unmarshal.go
```diff
@@ -253,15 +220,50 @@ func set(dest reflect.Value, argNameWords []string, value string) error {
...
-		field := dest.FieldByName(fieldName)
-		if !field.IsValid() {
-			return &UnknownArgError{}
+		if fieldIndex, exist := fieldIndexByName[fieldName]; exist {
+			return set(dest.Field(fieldIndex), argNameWords[1:], value)
+		}
+
+		//  If it do not exist we try to find it in nested anonymous field
+		for i := len(anonymousFieldIndex) - 1; i >= 0; i-- {
+			err := set(dest.Field(anonymousFieldIndex[i]), argNameWords, value)
```
loicbourgois: can't we first check the type of the field, instead of blindly trying all the anonymous fields ?

internal/args/unmarshal.go
```diff
@@ -253,15 +220,50 @@ func set(dest reflect.Value, argNameWords []string, value string) error {
...
 
-		// try to find the correct field in the struct.
+		// We cannot rely on dest.GetFieldByName() as reflect library is doing deep traversing when using anonymous field.
+		// Because of that we should rely on our own logic
+		//
+		// - First we try to find a field with the correct name in the current struct
+		// - If it does not exist we try to find it in all nested anonymous field
+		//   Anonymous fields are traversed from last to first as the last one in the struct declaration should take precedence
+
+		// We construct two cache:
```
QuentinBrosse: caches

internal/args/unmarshal.go
```diff
@@ -253,15 +220,50 @@ func set(dest reflect.Value, argNameWords []string, value string) error {
...
+		anonymousFieldIndex := []int(nil)
```
QuentinBrosse: anonymousFieldIndexes

internal/args/unmarshal.go
```diff
@@ -253,15 +220,50 @@ func set(dest reflect.Value, argNameWords []string, value string) error {
 			return &MissingStructFieldError{Dest: dest.Interface()}
 		}
 
-		// try to find the correct field in the struct.
+		// We cannot rely on dest.GetFieldByName() as reflect library is doing deep traversing when using anonymous field.
```
QuentinBrosse: Do we want to move this code in our own `fieldByName` function?

internal/args/unmarshal.go
```diff
@@ -253,15 +220,50 @@ func set(dest reflect.Value, argNameWords []string, value string) error {
...
+		//  If it do not exist we try to find it in nested anonymous field
+		for i := len(anonymousFieldIndex) - 1; i >= 0; i-- {
+			err := set(dest.Field(anonymousFieldIndex[i]), argNameWords, value)
+			switch err.(type) {
+			case nil:
+				// If we got no error the field was correctly set we return nil.
+				return nil
+			case *UnknownArgError:
+				// If err is an UnknownArgError this could mean the field in in another anonymous field
+				// we continue to the anonymous field
```
QuentinBrosse: We continue to the previous anonymous field.

internal/args/unmarshal.go
jerome-quere: no, because this needs to be recursive as an anonymousField may alson have anonymousField and so on,.

internal/args/unmarshal.go
jerome-quere: This is not just a get by field as it should create all pointer along the way this is exactly what set is for.

PR #651, opened 2020-01-23
chore: add validation func on organization arg-spec

loicbourgois:


internal/core/validate.go
```diff
@@ -118,6 +118,16 @@ func ValidateSecretKey() ArgSpecValidateFunc {
 }
 
 func ValidateOrganisationID() ArgSpecValidateFunc {
```
kindermoumoute: ValidateOrganizationID

internal/core/validate.go
```diff
@@ -118,6 +118,16 @@ func ValidateSecretKey() ArgSpecValidateFunc {
...
+	return func(argSpec *ArgSpec, valueI interface{}) error {
+		value := valueI.(string)
+		if value == "" {
+			return nil
+		}
+		return ValidateOrganisationIDRequired()(argSpec, valueI)
+	}
+}
+
+func ValidateOrganisationIDRequired() ArgSpecValidateFunc {
```
kindermoumoute: can you comment: why is this so complicated to validate organization ID? Why do we need 2 functions? Can't we just validate a UUID?

jerome-quere:
Di you check all the places `ValidateOrganisationID` was used before renaming method ?

internal/core/validate.go
kindermoumoute: Why is this called required? This arg spec is never required.

PR #652, opened 2020-01-23
chore: sort namespace in scw usage

jerome-quere:


cmd/scw/main.go
```diff
@@ -25,11 +25,12 @@ var (
 
 func main() {
 	// Import all commands available in CLI from various packages.
+	// NB: Merge order impacts scw usage sort.
```
loicbourgois: We should add a test with golden for `scw -h`

jerome-quere: This would mean copy pasting this list of merge. I don't see a good way of doing it.

loicbourgois: is a function `getCommands()` overkill ? 
Should the `main`directly do the merging ?

PR #654, opened 2020-01-24
feat(instance): add examples and test more examples

kindermoumoute:


internal/namespaces/instance/v1/custom_server_create.go
```diff
@@ -108,7 +108,29 @@ func serverCreateCommand() *core.Command {
...
+			{
+				Short:   "Create and start an instance on Ubuntu Bionic",
+				Request: `{"image":"ubuntu_bionic","start":true}`,
+			},
+			{
+				Short:   "Create a GP1-XS instance, give it a name and add tags",
+				Request: `{"image":"ubuntu_bionic","type":"GP1-XS","name":"foo","tags":["prod","blue"]}`,
+			},
+			{
+				Short:   "Create an instance with 2 additional block volumes (50GB and 100GB)",
```
QuentinBrosse: We could add `and a default root volume`

internal/namespaces/instance/v1/custom_server_create_test.go
```diff
@@ -43,33 +54,154 @@ func Test_CreateServer(t *testing.T) {
...
+				assert.Equal(t, instance.ServerStateRunning, ctx.Result.(*instance.Server).State)
+			},
+		}))
+
 		t.Run("With bootscript", core.Test(&core.TestConfig{
 			Commands:  GetCommands(),
-			Cmd:       "scw instance server create image=ubuntu-bionic bootscript-id=eb760e3c-30d8-49a3-b3ad-ad10c3aa440b",
+			Cmd:       "scw instance server create image=ubuntu_bionic bootscript-id=eb760e3c-30d8-49a3-b3ad-ad10c3aa440b",
 			AfterFunc: deleteServerAfterFunc,
 			Check: func(t *testing.T, ctx *core.CheckFuncCtx) {
```
QuentinBrosse: Why not using [TestCheckEqual](https://github.com/scaleway/scaleway-cli/blob/v2/internal/core/testing.go#L313)

kindermoumoute: Because this is typed

internal/namespaces/instance/v1/custom_server_create.go
kindermoumoute: It sounds confusing to write this because it is the default behavior for all the above commands as well.

QuentinBrosse: I can't see where it is documented for our users. Maybe it is the good PR to fille the long usage (cf l.42)

internal/namespaces/instance/v1/custom_server_create_test.go
QuentinBrosse: So could do something with TestCheckEqual, remove / improve it

kindermoumoute: `TestCheckEqual` is able to compare value between the result and an external object added in the Meta from the `BeforeFunc`, that's its only advantage IMO

internal/namespaces/instance/v1/custom_server_create.go
kindermoumoute: Yes, all custom instance commands deserve a second pass on the documentation, but this is not the scope of this PR

internal/namespaces/instance/v1/custom_server_create_test.go
jerome-quere: I agree typed tests make more sense. Let see if TestCheckEqual is useful in some future case.

PR #661, opened 2020-01-28
fix(human): fix support for stringer in human marshaler

jerome-quere:


internal/human/marshal_test.go
```diff
@@ -38,6 +40,11 @@ type Human struct {
 	Acquaintances []*Acquaintance
 }
 
+type Stringer struct{}
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion
type Stringer struct{}

```

internal/human/marshal.go
```diff
@@ -106,16 +105,20 @@ func marshalStruct(value reflect.Value, opt *MarshalOpt) (string, error) {
...
+		case isInterfaceNil(value.Interface()):
+			return nil, nil
+
 		// If data has a registered MarshalerFunc call it.
 		case marshalerFuncs[rType] != nil:
 			str, err := marshalerFuncs[rType](value.Interface(), subOpts)
 			return [][]string{{strings.Join(keys, "."), str}}, err
 
+		// If data is a stringers
+		case rType.Implements(reflect.TypeOf((*fmt.Stringer)(nil)).Elem()):
```
loicbourgois: shouldn't we use a custom `marshallerFunc` ?

jerome-quere: This is an interface not a type so the marshal func would not work.

internal/human/utils.go
```diff
@@ -9,3 +12,17 @@ func Capitalize(s string) string {
 	}
 	return ""
 }
+
+func isInterfaceNil(data interface{}) bool {
```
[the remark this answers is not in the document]
jerome-quere: A struct cannot be nil (a pointer to a string can)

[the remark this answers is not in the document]
QuentinBrosse: My bad

PR #667, opened 2020-02-04
feat(instance): Add examples in all instances commands

remyleone:


internal/namespaces/instance/v1/custom_security_group.go
```diff
@@ -324,6 +330,36 @@ func securityGroupUpdateCommand() *core.Command {
...
+		Examples: []*core.Example{
+			{
+				Short:   "Change the name of the given security group",
+				Request: `{"security_group_id": "da1550ea-af6c-4441-9ddc-7565c5e8a7dc", "name": "foobar"}`,
+			},
+			{
+				Short:   "Change the description of the given security group",
+				Request: `{"security_group_id": "69e03584-613b-49b9-87ec-a57d540a91d6", "description": "foobar"}`,
+			},
+			{
```
kindermoumoute: Do we need both `Enable stateful security group` and `Disable stateful security group`?

internal/namespaces/instance/v1/custom_security_group.go
```diff
@@ -324,6 +330,36 @@ func securityGroupUpdateCommand() *core.Command {
...
+				Short:   "Enable stateful security group",
+				Request: `{"security_group_id": "ceadfe27-d186-4c50-bc4f-ff84f14cce4d", "stateful": true}`,
+			},
+			{
+				Short:   "Disable stateful security group",
+				Request: `{"security_group_id": "2d5c6dfb-3e1d-4642-be57-906768122df8", "stateful": false}`,
+			},
+			{
+				Short:   "Set the default inbound policy as drop",
```
kindermoumoute: Move default inbound/outbound policy example at the top of this list

internal/namespaces/instance/v1/custom_security_group.go
```diff
@@ -243,6 +243,12 @@ func securityGroupClearCommand() *core.Command {
 		Resource:  "security-group",
 		Verb:      "clear",
 		ArgsType:  reflect.TypeOf(instanceResetSecurityGroupArgs{}),
+		Examples: []*core.Example{
+			{
+				Short:   "Remove all rules of the given security group",
+				Request: `{"security_group_id": "f8a5139c-6487-492d-a402-a9b9ffa21602"}`,
```
loicbourgois: All UUID used for examples should be 1111111-1111-1111-1111-111111111111
It makes it easier to notice this is an example and should be replaced to work properly.

PR #678, opened 2020-02-07
fix(core): bootstrap does not return errors

kindermoumoute:


internal/namespaces/instance/v1/custom_server.go
```diff
@@ -452,7 +452,7 @@ func serverDeleteCommand() *core.Command {
 			}
 
 			var multiErr error
-			if args.DeleteIP && server.Server.PublicIP != nil {
+			if args.DeleteIP && server.Server.PublicIP != nil && server.Server.PublicIP.Dynamic == false {
```
golangcibot: S1002: should omit comparison to bool constant, can be simplified to `!server.Server.PublicIP.Dynamic` (from `gosimple`)

golangcibot: S1002: should omit comparison to bool constant, can be simplified to `!server.Server.PublicIP.Dynamic` (from `gosimple`)

PR #680, opened 2020-02-10
chore: only use TestCheckCombine() when necessary

loicbourgois:


internal/core/build_info_test.go
```diff
@@ -33,9 +33,7 @@ func Test_CheckVersion(t *testing.T) {
 		},
 		BeforeFunc: deleteLatestVersionUpdateFile,
 		Cmd:        "scw plop",
-		Check: TestCheckCombine(
-			TestCheckStderrGolden(),
```
kindermoumoute: This test and all above tests should also have `TestCheckExitCode(0)`

PR #682, opened 2020-02-11
feat(autocomplete): support linux

kindermoumoute:


internal/namespaces/autocomplete/autocomplete.go
```diff
@@ -64,6 +64,7 @@ var autocompleteScripts = map[string]autocompleteScript{
 		CompleteScript: `eval "$(scw autocomplete script shell=bash)"`,
 		ShellConfigurationFile: map[string]string{
 			"darwin": path.Join(os.Getenv("HOME"), ".bash_profile"),
```
loicbourgois: maybe we should use `.bashrc` for macOS too ? 

I have both `.bashrc` and `.bash_profile`

Reading this, https://apple.stackexchange.com/a/51038, I'm not sure which is best.

QuentinBrosse: `.bash_profile` is widely used in macOS, we should use it.

PR #665, opened 2020-01-31
feat(core): print a warning when new release is available

kindermoumoute:

PR #685, opened 2020-02-11
fix(core): make warning the default logger level

remyleone:


kindermoumoute:
I reverted few changes made in https://github.com/scaleway/scaleway-cli/pull/665 and fixed the goldens

PR #689, opened 2020-02-11
feat(core): rename send_usage into send_telemetry

QuentinBrosse:


internal/namespaces/init/init.go
```diff
@@ -191,22 +191,22 @@ func initCommand() *core.Command {
 			}
 
 			// Ask for send usage permission
-			if args.SendUsage == nil {
+			if args.SendTelemetry == nil {
 				_, _ = interactive.Println()
 				_, _ = interactive.PrintlnWithoutIndent(`
-					To improve this tool we rely on diagnostic and usage data.
-					Sending such data is optional and can be disable at any time by running "scw config set send_usage false"
+					To improve this tools we rely on diagnostic and usage data.
```
kindermoumoute: should be `this tool`, meaning the CLI

QuentinBrosse: A rabase, fail...

PR #692, opened 2020-02-11
feat(args): add support for empty slice un/marshahaling

jerome-quere:


internal/args/marshal_test.go
```diff
@@ -101,6 +101,17 @@ func TestMarshal(t *testing.T) {
 		},
 	}))
 
+	t.Run("empty-slice", run(TestCase{
+		data: &Slice{
+			Strings:    []string{},
+			StringsPtr: []*string{},
```
kindermoumoute: StringsPtr should be `*[]string{}`

internal/args/unmarshal.go
```diff
@@ -177,14 +207,19 @@ func set(dest reflect.Value, argNameWords []string, value string) error {
 
 	case reflect.Slice:
 		// If type is a slice:
-		// We check if argNameWords[0] is an number to handle cases like keys.0.value=12
+
+		// We handle special case where array=none creates an empty slice.
```
QuentinBrosse: nit: `When array=none, creates an empty slice.`

internal/args/marshal_test.go
```diff
@@ -101,6 +101,17 @@ func TestMarshal(t *testing.T) {
...
+		},
+		expected: []string{
+			"strings=none",
```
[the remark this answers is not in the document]
jerome-quere: not extactly the same
`absent` => `[]string(nil)`
`=none` => `[]string{}`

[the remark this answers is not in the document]
kindermoumoute: AFAIK we don't have any use case for this as we always handle `[]string(nil) == []string{}` in our APIs

internal/args/marshal_test.go
```diff
@@ -104,11 +110,13 @@ func TestMarshal(t *testing.T) {
 	t.Run("empty-slice", run(TestCase{
 		data: &Slice{
 			Strings:    []string{},
+			SlicePtr:   &emptySlicePtr,
 			StringsPtr: []*string{},
```
kindermoumoute: IMO we shouldn't keep `[]*string{}` in this test

PR #693, opened 2020-02-11
fix(autocomplete): complete on fish subprocess

kindermoumoute:
This fix will work after fish 3.1, see https://github.com/fish-shell/fish-shell/issues/6214

### DIFF
--- a/internal/namespaces/autocomplete/autocomplete.go
+++ b/internal/namespaces/autocomplete/autocomplete.go
@@ -83,7 +83,7 @@ var autocompleteScripts = map[string]autocompleteScript{
 		CompleteFunc: `
 			complete --erase --command scw;
 			complete --command scw --no-files;
-			complete --command scw --arguments '(scw autocomplete complete fish -- (commandline) (commandline --cursor) (commandline --current-token) (commandline --tokenize --cut-at-cursor))';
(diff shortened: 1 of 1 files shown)


internal/namespaces/autocomplete/autocomplete.go
```diff
@@ -83,7 +83,7 @@ var autocompleteScripts = map[string]autocompleteScript{
 		CompleteFunc: `
 			complete --erase --command scw;
 			complete --command scw --no-files;
-			complete --command scw --arguments '(scw autocomplete complete fish -- (commandline) (commandline --cursor) (commandline --current-token) (commandline --tokenize --cut-at-cursor))';
+			complete --command scw --arguments '(scw autocomplete complete fish -- (commandline) (commandline --cursor) (commandline --current-token) (commandline --current-process --tokenize --cut-at-cursor))';
```
jerome-quere: do we need `--current-process` flag  for other  `commandline` command ?

kindermoumoute: `(commandline)` and `(commandline --cursor)` are not used yet. If we want to use them they must remain as is because the `cursor` points to the character index on the whole buffer ( returned by `commandline`)

PR #697, opened 2020-02-13
fix: properly handle maps in unmarshaller

loicbourgois:


internal/args/unmarshal_test.go
```diff
@@ -331,6 +331,38 @@ func TestUnmarshalStruct(t *testing.T) {
...
+			"name=coucou",
+			"elements.0.id=1453",
+			"elements.0.name=bob",
+		},
+		expected: &RecursiveWithMapOfRecursive{
+			Name: "coucou",
+			Elements: map[string]*RecursiveWithMapOfRecursive{
+				"0": {
+					ID:   1453,
+					Name: "bob",
```
loicbourgois: Only last field is properly set. 

Here, seems `id=1453` is erased.

internal/args/unmarshal_test.go
```diff
@@ -331,6 +331,38 @@ func TestUnmarshalStruct(t *testing.T) {
 			All: "all",
 		},
 	}))
+
+	t.Run("recursive-with-map-of-recursive-with-one-field-set", run(TestCase{
+		args: []string{
+			"name=coucou",
+			"elements.0.name=bob",
```
kindermoumoute: should also test `elements.0.elements.plop.name=world`

PR #698, opened 2020-02-13
feat(instance): update server volume ids

kindermoumoute:


internal/core/arg_specs.go
```diff
@@ -20,6 +24,21 @@ func (s *ArgSpecs) DeleteByName(name string) {
 			return
 		}
 	}
+	panic(fmt.Errorf("in DeleteByName: %s not found"))
```
golangcibot: SA5009: Printf format %s reads arg #1, but call has only 0 args (from `staticcheck`)

internal/core/arg_specs.go
```diff
@@ -20,6 +24,21 @@ func (s *ArgSpecs) DeleteByName(name string) {
...
+		if spec.Name == name {
+			newSpecs := ArgSpecs(nil)
+			newSpecs = append(newSpecs, (*s)[:i]...)
+			newSpecs = append(newSpecs, argSpec)
+			newSpecs = append(newSpecs, (*s)[i:]...)
+			*s = newSpecs
+			return
+		}
+	}
+	panic(fmt.Errorf("in AddBefore: %s not found"))
```
golangcibot: SA5009: Printf format %s reads arg #1, but call has only 0 args (from `staticcheck`)

PR #711, opened 2020-02-17
doc: release process

QuentinBrosse:


jerome-quere:
We should also consider pushing an image to docker hub with 
```
docker buildx build --platform=linux/amd64,linux/arm/7,linux/arm64,linux/i386 -t scaleway/scaleway-cli:yourawesomev2tag . --push
```

PR #719, opened 2020-02-24
chore: hide column 'valid until' in marketplace list

loicbourgois:


internal/namespaces/marketplace/v1/updated_commands.go
```diff
@@ -17,6 +17,7 @@ func updateCommands(commands *core.Commands) {
 }
 
 // customMarketplaceListImagesRun is the custom Run for scw marketplace list images
```
QuentinBrosse: nit: `updateMarketplaceListImages is the...`

PR #721, opened 2020-02-24
chore: list valid args when passing mal-formated argument

loicbourgois:


internal/core/cobra_utils_test.go
```diff
@@ -0,0 +1,48 @@
...
+			},
+		},
+	)
+}
+
+func Test_handleUnmarshalErrors(t *testing.T) {
+	t.Run("underscore", Test(&TestConfig{
+		Commands: testGetCommands(),
+		Cmd:      "scw test name_id",
+		Check: TestCheckCombine(
```
jerome-quere: I don't think golden are appropriate for this kind of test.
This should be somtehing like
```
Check: TestCheckError(&CliError{
   ...
}
```

PR #727, opened 2020-02-26
feat(instance): add server wait command

kindermoumoute:


internal/namespaces/instance/v1/custom_server.go
```diff
@@ -480,6 +480,27 @@ func serverRebootCommand() *core.Command {
 	}
 }
 
+func serverWaitCommand() *core.Command {
+	return &core.Command{
+		Short:     `Wait for server to reach a stable state`,
+		Long:      `Wait for server to reach a stable state. This command is equivalent to using the --wait flag when performing an action on a server.`,
```
loicbourgois: It's not really equivalent. 

`instance server wait` and `instance server update --wait` are not doing the same thing: the first does nothing, while the second updates a server. Both are blocking.

Equivalencies are `instance server update --wait` and `instance server update ; instance server wait`. 

> Wait for server to reach a stable state. Similar to using `--wait` flag, but without requiring a new action on the server.

internal/namespaces/instance/v1/custom_server.go
```diff
@@ -480,6 +480,27 @@ func serverRebootCommand() *core.Command {
 	}
 }
 
+func serverWaitCommand() *core.Command {
```
loicbourgois: Can `wait` fucntions be tested ? 

If so, please add a test.

internal/namespaces/instance/v1/custom_server.go
```diff
@@ -480,6 +480,27 @@ func serverRebootCommand() *core.Command {
...
+	return &core.Command{
+		Short:     `Wait for server to reach a stable state`,
+		Long:      `Wait for server to reach a stable state. This command is equivalent to using the --wait flag when performing an action on a server.`,
+		Namespace: "instance",
+		Resource:  "server",
+		Verb:      "wait",
+		ArgsType:  reflect.TypeOf(instanceActionRequest{}),
+		Run: func(ctx context.Context, argsI interface{}) (i interface{}, err error) {
+			return waitForServerFunc()(ctx, argsI, nil)
```
loicbourgois: The third argument does not seem to be used anywhere. 

Is it really necessarry to keep it ?

PR #731, opened 2020-02-27
fix(instance): use zone field in listing

kindermoumoute:


internal/namespaces/instance/v1/custom_image.go
```diff
@@ -74,17 +74,9 @@ func imageListBuilder(c *core.Command) *core.Command {
...
-					parsedZone, err := scw.ParseZone(getServerResponse.Server.Location.ZoneID)
-					if err != nil {
-						return nil, err
-					}
-					zone = parsedZone
-				}
 				newCustomImage.ServerID = getServerResponse.Server.ID
 				newCustomImage.ServerName = getServerResponse.Server.Name
-				newCustomImage.Zone = zone
+				newCustomImage.Zone = image.Zone
```
loicbourgois: can't we set it when creating `newCustomImage`, at line 65 ?

internal/namespaces/instance/v1/testdata/test-create-server-simple-default.stdout.golden
```diff
@@ -20,6 +20,7 @@ image.organization        51b656e3-4865-41e8-adbc-0c45bdd780db
 image.public              true
 image.root-volume         dd5f5c10-23b1-4c9c-8445-eb6740957c84
 image.state               available
+image.zone
```
loicbourgois: why is there no zone in create, but a zone is set in get ?

internal/namespaces/instance/v1/testdata/test-create-server-simple-default.stdout.golden
```diff
@@ -33,3 +34,4 @@ security-group.id         49822f6c-1d32-465c-a017-720691d159b2
 security-group.name       Default security group
 state-detail              
 arch                      x86_64
+zone
```
loicbourgois: why is there no zone in create, but a zone is set in get ?

internal/namespaces/instance/v1/testdata/test-list-server-simple.stdout.golden
```diff
@@ -1,12 +1,12 @@
...
+93eb4197-46ae-4c81-b6cf-1222a861edd2  cli-srv-upbeat-burnell        archived        51.158.119.154                 Ubuntu Bionic Beaver  []    few seconds ago    few seconds ago  d4067cdc-dc9d-4810-8a26-0dae51d7df42  false      1        49822f6c-1d32-465c-a017-720691d159b2  Default security group                  x86_64
+9424a299-8cf3-4125-9471-0d10d4df0412  cli-srv-hopeful-jennings      archived        163.172.177.30                 Ubuntu Bionic Beaver  []    few seconds ago    few seconds ago  d4067cdc-dc9d-4810-8a26-0dae51d7df42  false      1        49822f6c-1d32-465c-a017-720691d159b2  Default security group                  x86_64
+e34e7ca4-b5aa-4522-9cba-d16122f041f6  cli-srv-hardcore-einstein     running         51.158.120.130  10.68.26.121   Ubuntu Bionic Beaver  []    few seconds ago    few seconds ago  f974feac-abae-4365-b988-8ec7d1cec10d  false      1        49822f6c-1d32-465c-a017-720691d159b2  Default security group  booting kernel  x86_64
```
loicbourgois: should zone really disappear ? 

this looks like a regression

internal/namespaces/instance/v1/testdata/test-create-server-simple-default.stdout.golden
kindermoumoute: The issue comes from the cassette because the field zone was not returned by the API until recently.

internal/namespaces/instance/v1/testdata/test-list-server-simple.stdout.golden
kindermoumoute: The issue comes from the cassette because the field zone was not returned by the API until recently.

PR #732, opened 2020-02-27
fix(instance): enhance server type listing

kindermoumoute:


internal/namespaces/instance/v1/custom_server_type.go
```diff
@@ -52,11 +53,20 @@ func serverTypeListBuilder(c *core.Command) *core.Command {
 		}
 
 		sort.Slice(serverTypes, func(i, j int) bool {
-			return serverTypes[i].Name < serverTypes[j].Name
+			catA := serverTypeCategory(serverTypes[i].Name)
```
loicbourgois: `catA` -> `categoryA`

cat means "chat"

PR #737, opened 2020-03-02
chore: improve human marshal for nil value

loicbourgois:


loicbourgois:
Maybe we should use `NA` instead of `-` ?

PR #738, opened 2020-03-02
chore: reorder instance server list collumns

loicbourgois:


internal/namespaces/instance/v1/custom_server.go
```diff
@@ -87,22 +88,23 @@ func serversMarshalerFunc(i interface{}, opt *human.MarshalOpt) (string, error)
 		humanServers = append(humanServers, &humanServerInList{
 			ID:                server.ID,
 			Name:              server.Name,
+			CommercialType:    server.CommercialType,
```
kindermoumoute: Is possible to rename this field to `Type` in order to be consistent with `server create`?

PR #742, opened 2020-03-04
doc: add a README for the script section

remyleone:


jerome-quere:
FYI: @remyleone I rename this PR as a `doc:` as this does not affect software. `feat` should be used when we add a feature that makes should increment `minor` version

PR #755, opened 2020-03-09
fix(core): fix parallel tests not being parallel

### LINKED ISSUE #764: Tests are not run in parallel
A missing `!` made all tests run in sequence. When adding the `!`, a lots of tests are failing. 
I'll try to sum up every tests here

Sh4d1:
Tests were not run in parallel :sweat_smile: 

Fixes #764 

Signed-off-by: Patrik Cyvoct <patrik@ptrk.io>

### DIFF
--- a/internal/human/marshal.go
+++ b/internal/human/marshal.go
@@ -45,14 +45,17 @@ func Marshal(data interface{}, opt *MarshalOpt) (string, error) {
 
 	rType := rValue.Type()
 
+	// safely get the marshalerFunc
+	marshalerFunc, _ := getMarshalerFunc(rType)
+
 	switch {
 	// If data is nil
 	case isInterfaceNil(data):
 		return defaultMarshalerFunc(nil, opt)
 
 	// If data has a registered MarshalerFunc call it
-	case marshalerFuncs[rType] != nil:
-		return marshalerFuncs[rType](rValue.Interface(), opt)
+	case marshalerFunc != nil:
+		return marshalerFunc(rValue.Interface(), opt)
 
 	// Handle special well known interface
 	case rType.Implements(reflect.TypeOf((*Marshaler)(nil)).Elem()):
@@ -108,14 +111,17 @@ func marshalStruct(value reflect.Value, opt *MarshalOpt) (string, error) {
 		}
 		rType := value.Type()
 
+		// safely get the marshaler func
+		marshalerFunc, _ := getMarshalerFunc(rType)
+
 		switch {
 		// If data is nil
 		case isInterfaceNil(value.Interface()):
 			return nil, nil
 
 		// If data has a registered MarshalerFunc call it.
-		case marshalerFuncs[rType] != nil:
-			str, err := marshalerFuncs[rType](value.Interface(), subOpts)
+		case marshalerFunc != nil:
+			str, err := marshalerFunc(value.Interface(), subOpts)
 			return [][]string{{strings.Join(keys, "."), str}}, err
 
 		// If data is a stringers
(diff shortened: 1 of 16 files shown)


Sh4d1:
Maybe disable it by defaut :thinking:

QuentinBrosse:
> Maybe disable it by default 🤔

Most of our tests should pass in parallel has we are using cassets.

internal/namespaces/instance/v1/custom_server_test.go
```diff
@@ -263,7 +266,8 @@ func Test_ServerDelete(t *testing.T) {
 			core.TestCheckGolden(),
 			core.TestCheckExitCode(0),
 		),
-		AfterFunc: core.ExecAfterCmd(`scw instance delete volume volume-id={{ (index .Server.Volumes "1").ID }}`),
+		AfterFunc:       core.ExecAfterCmd(`scw instance delete volume volume-id={{ (index .Server.Volumes "1").ID }}`),
```
remyleone: It should be `scw instance volume delete`

internal/namespaces/instance/v1/custom_server_test.go
```diff
@@ -252,7 +254,8 @@ func Test_ServerDelete(t *testing.T) {
 			core.TestCheckGolden(),
 			core.TestCheckExitCode(0),
 		),
-		AfterFunc: core.ExecAfterCmd(`scw instance delete volume volume-id={{ (index .Server.Volumes "0").ID }}`),
+		AfterFunc:       core.ExecAfterCmd(`scw instance delete volume volume-id={{ (index .Server.Volumes "0").ID }}`),
```
remyleone: It should be `scw instance volume delete`

internal/human/marshal.go
```diff
@@ -45,14 +45,17 @@ func Marshal(data interface{}, opt *MarshalOpt) (string, error) {
 
 	rType := rValue.Type()
 
+	// safely get the marshalerFunc
+	marshalerFunc, _ := getMarshalerFunc(rType)
```
loicbourgois: why do we work on `marshalerFunc` ? 

is this necessary to fix the problem of parallel tests ? 

shouldn't this be in another PR ?

Sh4d1: Changing the map to a sync map means changing the way we access the map, thus the getter :)

internal/namespaces/instance/v1/custom_server_create_test.go
```diff
@@ -230,6 +230,7 @@ func Test_CreateServer(t *testing.T) {
 }
 
 // None of the tests below should succeed to create an instance.
+// these tests need to be run in sequence
```
loicbourgois:
```suggestion
// these tests need to be run in sequence
```
```suggestion
// These tests need to be run in sequence.
```

internal/namespaces/instance/v1/custom_server_test.go
```diff
@@ -223,14 +223,15 @@ func Test_ServerUpdateCustom(t *testing.T) {
...
-				core.ExecAfterCmd(`scw instance delete volume volume-id={{ (index .Server.Volumes "1").ID }}`),
+				core.ExecAfterCmd(`scw instance volume delete volume-id={{ (index .Server.Volumes "0").ID }}`),
+				core.ExecAfterCmd(`scw instance volume delete volume-id={{ (index .Server.Volumes "1").ID }}`),
 				deleteServer("Server"),
 			),
 		}))
 	})
 }
 
+// these tests needs to be run in sequence
```
loicbourgois:
```suggestion
// These tests needs to be run in sequence.
```

internal/namespaces/instance/v1/custom_server_create_test.go
loicbourgois: Why ? 

Usually, comments should explain `why`, not `what`.

internal/namespaces/instance/v1/custom_server_test.go
loicbourgois: Why ? 

Comments should explain `why`, not `what`.

internal/namespaces/instance/v1/custom_server_create_test.go
Sh4d1: Good question :sweat_smile: it doesnt work in parallel but works in sequence. Not sure why :thinking:

internal/namespaces/instance/v1/custom_server_test.go
Sh4d1: Good question :sweat_smile: it doesnt work in parallel but works in sequence. Not sure why :thinking:

internal/namespaces/instance/v1/custom_server_create_test.go
Sh4d1: @loicbourgois fixed

PR #756, opened 2020-03-10
doc: add sudo to chmod command during install

angristan:
Since we download the binary with sudo, we need sudo to chmod +x it.

---
From a fellow Scaleway member 🤓

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,7 +70,7 @@ echo $PATH
 /usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
 
 # Download the release from github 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -73,7 +73,7 @@ echo $PATH
 sudo curl -o /usr/local/bin/scw -L "https://github.com/scaleway/scaleway-cli/releases/download/v2.0.0-beta.1/scw-v2.0.0-beta.1-darwin-x86_64"
```
QuentinBrosse: We don't need `sudo` at all for macOS. At least for standard macOS installations.
Can you remove it in `curl` and `chmod` please? :)

angristan:
squashed and force pushed my commits

README.md
angristan: Indeed 👍

angristan:
Ok, good to know 🙏

PR #770, opened 2020-03-12
chore: add github actions

QuentinBrosse:


.github/workflows/unit-tests.yaml
```diff
@@ -0,0 +1,51 @@
...
+#on:
+#  push:
+#    branches: [ v2 ]
+#  pull_request:
+
+on:
+  pull_request:
+
+jobs:
+  ubuntu-unit-test:
```
Sh4d1:
```suggestion
different-os-unit-test
```

.github/workflows/unit-tests.yaml
```diff
@@ -0,0 +1,48 @@
+name: Unit Tests
+
+on:
+  push:
+    branches: [ v2 ]
```
Sh4d1: Running tests once merged? :thinking:

.github/workflows/unit-tests.yaml
```diff
@@ -0,0 +1,46 @@
+name: Unit Tests
```
jerome-quere: These are not unit test as it's a mix of unit and integration.

.github/workflows/unit-tests.yaml
```diff
@@ -0,0 +1,46 @@
...
+
+on:
+  pull_request:
+
+jobs:
+  os-unit-test:
```
jerome-quere: same remark here

.github/workflows/unit-tests.yaml
```diff
@@ -0,0 +1,46 @@
...
+      - name: Install Go
+        uses: actions/setup-go@v1
+        with:
+          go-version: ${{ matrix.go-version }}
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 1
+      - name: Build binary
+        run: GOARCH=${{ matrix.arch }} go build ./...
```
jerome-quere: - Should we build cmd/scw ?
- Shoud we use build script that already do cross compilation ?

PR #774, opened 2020-03-14
feat(k8s): add wait and status color to k8s node

Sh4d1:

### DIFF
--- a/internal/namespaces/k8s/v1beta4/custom_node.go
+++ b/internal/namespaces/k8s/v1beta4/custom_node.go
@@ -15,10 +15,6 @@ const (
 	nodeActionTimeout = 10 * time.Minute
 )
 
-//
-// Marshalers
-//
-
 // nodeStatusMarshalerFunc marshals a k8s.NodeStatus.
 var (
 	nodeStatusAttributes = human.Attributes{
--- a/internal/namespaces/k8s/v1beta4/custom_node.go
+++ b/internal/namespaces/k8s/v1beta4/custom_node.go
@@ -1,0 +1,57 @@
+package k8s
+
+import (
+	"context"
+	"time"
+
+	"github.com/fatih/color"
+	"github.com/scaleway/scaleway-cli/internal/core"
+	"github.com/scaleway/scaleway-cli/internal/human"
+	k8s "github.com/scaleway/scaleway-sdk-go/api/k8s/v1beta4"
+	"github.com/scaleway/scaleway-sdk-go/scw"
+)
+
+const (
+	nodeActionTimeout = 10 * time.Minute
+)
+
+//
+// Marshalers
+//
+
+// nodeStatusMarshalerFunc marshals a k8s.ClusterStatus.
+var (
+	nodeStatusAttributes = human.Attributes{
+		k8s.NodeStatusCreating:      color.FgBlue,
+		k8s.NodeStatusRebooting:     color.FgBlue,
+		k8s.NodeStatusReady:         color.FgGreen,
+		k8s.NodeStatusNotready:      color.FgYellow,
+		k8s.NodeStatusCreationError: color.FgRed,
+		k8s.NodeStatusLocked:        color.FgRed,
+		k8s.NodeStatusWarning:       color.FgHiYellow,
+	}
+)
+
+const (
+	nodeActionReboot = iota
+)
+
+func nodeRebootBuilder(c *core.Command) *core.Command {
+	c.WaitFunc = waitForNodeFunc(nodeActionReboot)
+	return c
+}
+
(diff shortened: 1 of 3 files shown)


internal/namespaces/k8s/v1beta4/custom_node.go
```diff
@@ -0,0 +1,57 @@
...
+
+	"github.com/fatih/color"
+	"github.com/scaleway/scaleway-cli/internal/core"
+	"github.com/scaleway/scaleway-cli/internal/human"
+	k8s "github.com/scaleway/scaleway-sdk-go/api/k8s/v1beta4"
+	"github.com/scaleway/scaleway-sdk-go/scw"
+)
+
+const (
+	nodeActionTimeout = 10 * time.Minute
```
golangcibot: `nodeActionTimeout` is unused (from `deadcode`)

internal/namespaces/k8s/v1beta4/custom_node.go
```diff
@@ -0,0 +1,57 @@
...
+)
+
+//
+// Marshalers
+//
+
+// nodeStatusMarshalerFunc marshals a k8s.ClusterStatus.
+var (
+	nodeStatusAttributes = human.Attributes{
+		k8s.NodeStatusCreating:      color.FgBlue,
```
loicbourgois: isn't this list of attributes similar to one used in another command ? 

can't we use a common list, to avoid code duplication ?

internal/namespaces/k8s/v1beta4/custom_node.go
```diff
@@ -0,0 +1,57 @@
...
+		k8s.NodeStatusReady:         color.FgGreen,
+		k8s.NodeStatusNotready:      color.FgYellow,
+		k8s.NodeStatusCreationError: color.FgRed,
+		k8s.NodeStatusLocked:        color.FgRed,
+		k8s.NodeStatusWarning:       color.FgHiYellow,
+	}
+)
+
+const (
+	nodeActionReboot = iota
```
loicbourgois: will there be other actions in the future ?

internal/namespaces/k8s/v1beta4/custom_node.go
```diff
@@ -0,0 +1,57 @@
...
+
+const (
+	nodeActionTimeout = 10 * time.Minute
+)
+
+//
+// Marshalers
+//
+
+// nodeStatusMarshalerFunc marshals a k8s.ClusterStatus.
```
loicbourgois: this comment doesn't make sense

internal/namespaces/k8s/v1beta4/custom_node.go
```diff
@@ -0,0 +1,57 @@
...
+	k8s "github.com/scaleway/scaleway-sdk-go/api/k8s/v1beta4"
+	"github.com/scaleway/scaleway-sdk-go/scw"
+)
+
+const (
+	nodeActionTimeout = 10 * time.Minute
+)
+
+//
+// Marshalers
```
loicbourgois: not everything under is to do with marshalers 

add other comments, or remove this one

internal/namespaces/k8s/v1beta4/custom_node.go
loicbourgois: we're starting to have a lot of different timeouts in different files, although they are used for the same kind of operations

maybe we should have a single file with base timeouts as `baseTimeout = x * unit`, and then reuse them in the different `custom_xxx.go` files with `nodeActionTimeout = baseTimeout` ?

internal/namespaces/k8s/v1beta4/custom_node.go
Sh4d1: Not really, there are some for cluster, pool and nodesm which are not the same

internal/namespaces/k8s/v1beta4/custom_node.go
Sh4d1:
```suggestion
nodeStatusMarshalerFunc marshals a k8s.NodeStatus.
```

internal/namespaces/k8s/v1beta4/custom_node.go
Sh4d1: there could be :sweat_smile:

internal/namespaces/k8s/v1beta4/custom_node.go
Sh4d1: There are used for the same kind of operations, but not the same operations. In the future they may have different values. I can set a basetimeout in the meantime though

Sh4d1: I'll open a new PR to set a basetimeout once this one is merged if you don't mind :)

PR #788, opened 2020-03-18
fix(core): disable check args exist valid for raw

### LINKED ISSUE #787: Autocomplete is not working on master
```
scw autocomplete complete zsh -- 2 scw
Unknown argument '2'

Hint:
Valid arguments are: 
```

This is because https://github.com/scaleway/scaleway-cli/blob/v2/internal/core/cobra_utils.go#L38 is executed even on rawargs

Sh4d1:
Fixes #787 


internal/core/cobra_utils.go
```diff
@@ -33,17 +33,19 @@ func cobraRun(ctx context.Context, cmd *Command) func(*cobra.Command, []string)
...
-			// TODO: handle args such as tags.index
-			if cmd.ArgSpecs.GetByName(arguments[0]) == nil &&
-				cmd.ArgSpecs.GetByName(arguments[0]+".{index}") == nil &&
-				!strings.Contains(arguments[0], ".") {
-				return handleUnmarshalErrors(cmd, &args.UnmarshalArgError{
-					Err:     &args.UnknownArgError{},
-					ArgName: arguments[0],
-				})
+		// Check args exist valid if ArgsType is not args.RawArgs
+		if reflect.TypeOf(args.RawArgs{}) != cmd.ArgsType {
```
loicbourgois:
```suggestion
		if cmd.ArgsType != reflect.TypeOf(args.RawArgs{}) {
```

nit: i find it easier to read

internal/namespaces/autocomplete/autocomplete_test.go
```diff
@@ -3,9 +3,91 @@ package autocomplete
...
 	"github.com/stretchr/testify/assert"
 )
 
+func Test_Autocomplete(t *testing.T) {
+	t.Run("zsh", func(t *testing.T) {
+		t.Run("Level1", core.Test(&core.TestConfig{
+			Commands: GetCommands(),
+			Cmd:      "scw autocomplete complete zsh -- 1",
+			Check: core.TestCheckCombine(
+				core.TestCheckGolden(),
```
loicbourgois: golden files should be used to test UI, not to test the content of a result 

can we have the response inlined here please ?

loicbourgois: also, don't forget to remove the unused golden files

internal/namespaces/autocomplete/testdata/test-autocomplete-bash-level1.stdout.golden
```diff
@@ -0,0 +1 @@
+-
```
loicbourgois: please remove unused goldens

internal/core/cobra_utils_test.go
```diff
@@ -71,6 +88,25 @@ func Test_handleUnmarshalErrors(t *testing.T) {
 	}))
 }
 
+func Test_RawArgs(t *testing.T) {
```
loicbourgois: shoudn't we have a `testcobraRun` func that only test `cobraRun` ? 

I find it strnage that we hav to test an entire command. 

Shouldn't we only call `cobraRun()` in the test.

Sh4d1: we could :smile: I'll fix that

Sh4d1: Though it may be more complicated :thinking:

internal/core/cobra_utils.go
```diff
@@ -33,17 +33,19 @@ func cobraRun(ctx context.Context, cmd *Command) func(*cobra.Command, []string)
...
-			// TODO: handle args such as tags.index
-			if cmd.ArgSpecs.GetByName(arguments[0]) == nil &&
-				cmd.ArgSpecs.GetByName(arguments[0]+".{index}") == nil &&
-				!strings.Contains(arguments[0], ".") {
-				return handleUnmarshalErrors(cmd, &args.UnmarshalArgError{
-					Err:     &args.UnknownArgError{},
-					ArgName: arguments[0],
-				})
+		// Check args exist valid if ArgsType is not args.RawArgs
+		if cmd.ArgsType != reflect.TypeOf(args.RawArgs{}) {
```
kindermoumoute: shouldn't the condition be `== reflect.TypeOf(args.RawArgs{})`? 🤔

Sh4d1: If we enter this condition with a raw args it will always throuh a unknown arg error since th argument does not have a name :sweat_smile:

PR #811, opened 2020-03-26
fix(core): fix autocomplete edge cases

### LINKED ISSUE #810: Autocomplete not working when using in shell functions or different binary name
For instance in zsh:
```zsh
test() {
  wHATEVER=foo scw "$@"
}
compdef -e 'WHATEVER=foo _scw' stg
```
breaks to autocompletion for test, which should be working.

Same thing happend if I change the binary name `scw` to another one.

Sh4d1:
Fixes #810 

- a test is removed since it's never used in real world: autocompletion is only called once the command in typed and a space is added

### DIFF
--- a/internal/core/autocomplete.go
+++ b/internal/core/autocomplete.go
@@ -2,7 +2,6 @@ package core
 
 import (
 	"context"
-	"os"
 	"regexp"
 	"sort"
 	"strconv"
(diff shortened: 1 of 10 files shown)


jerome-quere:
Instead of replacing hardcoded `scw` by `os.Args[0]` it could be a better idea to make `BinaryName` a parameter in the `BootstrapConfig` this config could be accessible through context where needed

internal/core/autocomplete.go
```diff
@@ -196,7 +197,7 @@ func (node *AutoCompleteNode) isLeafCommand() bool {
 // BuildAutoCompleteTree builds the autocomplete tree from the commands, subcommands and arguments
 func BuildAutoCompleteTree(commands *Commands) *AutoCompleteNode {
 	root := NewAutoCompleteCommandNode()
-	scwCommand := root.GetChildOrCreate("scw")
+	scwCommand := root.GetChildOrCreate(os.Args[0])
```
jerome-quere, line 200: Yes But... we should not realy on `os.Args[0]` as this is a global varibale.
The binary name should be a parameter in BootstrapConfig and passed allong with context,

PR #815, opened 2020-03-27
test: improve error message when using TestCheckCombine with only one test

jerome-quere:


QuentinBrosse:
We also have this type of message in `AfterFuncCombine` and `BeforeFuncCombine`.

jerome-quere:
@QuentinBrosse  I don't have access to `t *testing.T` in After/Before func this would require a big refacto to be able to do this

internal/core/testing.go
```diff
@@ -330,7 +330,7 @@ func ExecAfterCmd(cmd string) AfterFunc {
 // TestCheckCombine combines multiple check functions into one.
 func TestCheckCombine(checks ...TestCheck) TestCheck {
 	return func(t *testing.T, ctx *CheckFuncCtx) {
-		assert.Equal(t, true, len(checks) > 1)
+		assert.Equal(t, true, len(checks) > 1, "You should not use TestCheckCombine with only one test")
```
QuentinBrosse: Just to keep consistency with `{Before|AfterFuncCombine}`:

```suggestion
		assert.Equal(t, true, len(checks) > 1, "TestCheckCombine must be used to combine more than one TestCheck")
```

PR #818, opened 2020-03-27
feat(init): rename send-telemetry arg and improve usage

QuentinBrosse:


internal/namespaces/init/init_test.go
```diff
@@ -0,0 +1,18 @@
...
+import (
+	"testing"
+
+	"github.com/scaleway/scaleway-cli/internal/core"
+)
+
+func Test_Init(t *testing.T) {
+	t.Run("Helper", core.Test(&core.TestConfig{
+		Commands: GetCommands(),
+		Cmd:      "scw init --help",
```
loicbourgois: usage should be tested with `Test_AllUsage`

but `init` has a var to build its help 

see line 1 in golden 
> Initialize the active profile of the config located in /Users/qbrosse/.config/scw/config.yaml 

maybe we need to overide  `$HOME`in the test

QuentinBrosse: It is [excluded](https://github.com/scaleway/scaleway-cli/blob/v2/cmd/scw/main_test.go#L24) from `Test_AllUsage` for this reason. :palm_face:

PR #827, opened 2020-04-02
feat(instance): add stocks in server-type list

QuentinBrosse:
This PR also proposes a new `EnumMarshalFunc` as a replacement of `BindAttributesMarshalFunc`.


internal/namespaces/instance/v1/custom_server_type.go
```diff
@@ -33,30 +47,39 @@ func serverTypeListBuilder(c *core.Command) *core.Command {
...
-
-		listServersTypesResponse := originalRes.(*instance.ListServersTypesResponse)
 		serverTypes := []*customServerType(nil)
 
-		for name, serverType := range listServersTypesResponse.Servers {
+		// Get server availabilities.
+		serverAvailabilities := make(map[string]instance.ServerTypesAvailability)
+		availabilitiesResponse, err := api.GetServerTypesAvailability(&instance.GetServerTypesAvailabilityRequest{})
+		if err != nil {
+			// Do nothing and show degraded results.
```
jerome-quere, line 77: We should be really careful with this type of optimization. While it's a excellent idea for human display it could be a source of many strange bugs when doing scripting and automation as the command will not return an error but will not contain the correct information? 

Can we consider this case will almost never happen and that user experience will almost never be affected by this use-case and at the same time guarantee automation?

QuentinBrosse: Good catch, what is your point? Do you prefer to return an error here? I agree.

jerome-quere: yes let's return an error

PR #833, opened 2020-04-03
docs: fix example with positional arg

kindermoumoute:


internal/core/cobra_usage_builder.go
```diff
@@ -91,6 +91,18 @@ func buildExamples(cmd *Command) string {
 				panic(fmt.Errorf("in command '%s', example '%s': %w", cmd.getPath(), cmdExample.Short, err))
 			}
 			var cmdArgsAsStrings, err = args.MarshalStruct(cmdArgs)
+			positionalArg := cmd.ArgSpecs.GetPositionalArg()
+			if positionalArg != nil {
+				for i, cmdArg := range cmdArgsAsStrings {
+					if !strings.HasPrefix(cmdArg, positionalArg.Prefix()) {
+						continue
+					}
+					cmdArgsAsStrings[i] = strings.TrimLeft(cmdArg, positionalArg.Prefix())
```
jerome-quere, line 100: Add a comment 
```
// Switch the positional args with args at position 0 to make sure positional args is always at the beginning
```

PR #834, opened 2020-04-03
feat(init): handle empty config

kindermoumoute:


jerome-quere:
This is not a feature as it's doesn't expose new feature. It's more a `vendor: update scaleway sdk`

PR #836, opened 2020-04-03
feat(init): ask to remove v1 config

kindermoumoute:


internal/namespaces/init/init.go
```diff
@@ -231,6 +233,24 @@ func initCommand() *core.Command {
 				args.InstallAutocomplete = scw.BoolPtr(installAutocomplete)
 			}
 
+			// Ask whether to remove v1 configuration file is it exists
+			homeDir, err := os.UserHomeDir()
```
jerome-quere, line 237: Should we make `getConfigV1FilePath` in SDK public ?

internal/namespaces/init/init.go
```diff
@@ -231,6 +233,24 @@ func initCommand() *core.Command {
...
+			if err == nil {
+				configPath := path.Join(homeDir, ".scwrc")
+				if _, err := os.Stat(configPath); err == nil {
+					removeV1ConfigFile, err := interactive.PromptBoolWithConfig(&interactive.PromptBoolConfig{
+						Prompt:       "Do you want to permanently remove old configuration file (" + configPath + ")?",
+						DefaultValue: true,
```
jerome-quere, line 243: Default to `no` especially because this what happen in non-interactive

internal/namespaces/init/init.go
```diff
@@ -65,6 +66,7 @@ type initArgs struct {
 	SendTelemetry       *bool
 	WithSSHKey          *bool
 	InstallAutocomplete *bool
+	V1ConfigFilePath    *string
```
jerome-quere, line 69: This is not an argument of the command.
If we really want to make it an arg we could do
`scw init delete-v1-config=true`
but i'm not a fan

internal/namespaces/init/init.go
```diff
@@ -231,6 +233,24 @@ func initCommand() *core.Command {
 				args.InstallAutocomplete = scw.BoolPtr(installAutocomplete)
 			}
 
+			// Ask whether to remove v1 configuration file is it exists
```
jerome-quere, line 236: I would move this whole portion to the end of the run

internal/namespaces/init/init.go
kindermoumoute: Not a big fan since it would encourage to use it

internal/namespaces/init/init.go
kindermoumoute: It is not consistent with other flags

internal/namespaces/init/init.go
```diff
@@ -104,6 +106,10 @@ func initCommand() *core.Command {
 				Name:  "install-autocomplete",
 				Short: "Whether the autocomplete script should be installed during initialisation",
 			},
+			{
+				Name:  "remove-old-config",
+				Short: "Whether the autocomplete script should be installed during initialisation",
```
jerome-quere, line 111: not the correct documentation

PR #840, opened 2020-04-06
refacto: add support for command interceptor

jerome-quere:

PR #842, opened 2020-04-06
refacto: handle sdk error with interecptor

jerome-quere:
Should be merged after #840

### DIFF
--- a/internal/core/command_interceptor.go
+++ b/internal/core/command_interceptor.go
@@ -70,7 +70,6 @@ func sdkStdErrorInterceptor(ctx context.Context, args interface{}, runner Comman
 		}
 
 	case *scw.QuotasExceededError:
-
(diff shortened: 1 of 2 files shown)


internal/core/command_interceptor.go
```diff
@@ -29,6 +33,84 @@ func combineCommandInterceptor(interceptors ...CommandInterceptor) CommandInterc
...
+			Err:     err,
+		}
+	case *scw.ResponseError:
+		return nil, &CliError{
+			Message: sdkError.Message,
+			Err:     sdkError,
+		}
+	case *scw.InvalidArgumentsError:
+		reasonsMap := map[string]string{
+			"unknown":    "is invalid for unexpected reason",
```
Sh4d1, line 50: `an unexpected reason` or `unexpected reasons` ?

PR #854, opened 2020-04-09
fix(instance): list image with image from not found server

kindermoumoute:


internal/namespaces/instance/v1/custom_image.go
```diff
@@ -115,6 +115,10 @@ func imageListBuilder(c *core.Command) *core.Command {
 				}
 				getServerResponse, err := api.GetServer(&serverReq)
 				if err != nil {
+					if _, ok := err.(*scw.ResourceNotFoundError); ok {
+						newCustomImage.ServerName = "-"
+						continue
```
Sh4d1, line 120: It will not be added in the array no?

internal/namespaces/instance/v1/custom_image.go
```diff
@@ -115,6 +115,10 @@ func imageListBuilder(c *core.Command) *core.Command {
 				}
 				getServerResponse, err := api.GetServer(&serverReq)
 				if err != nil {
+					if _, ok := err.(*scw.ResourceNotFoundError); ok {
```
Sh4d1, line 118: iirc this does not work

Sh4d1: In fact it does not work :sweat_smile:

PR #897, opened 2020-04-22
feat(instance): add console command

Sh4d1:


internal/gotty/client.go
```diff
@@ -0,0 +1,208 @@
...
+	"syscall"
+	"time"
+
+	"github.com/containerd/console"
+	"github.com/gorilla/websocket"
+	"github.com/scaleway/scaleway-sdk-go/scw"
+)
+
+// Scaleway GoTTY websocket endpoint
+var wsURLs = map[scw.Zone]string{
```
Sh4d1, line 18: Not sure which is better between `tty.scaleway.com` and `tty-<region>.scaleway.com` :thinking:

internal/gotty/client.go
```diff
@@ -0,0 +1,208 @@
...
+	err = cns.SetRaw()
+	if err != nil {
+		return fmt.Errorf("error setting raw terminal: %w", err)
+	}
+	defer cns.Reset()
+	defer cns.Close()
+
+	// Create a chanel that will receive all resizes signals
+	resizeChan := make(chan os.Signal, 1)
+	signal.Notify(resizeChan, syscall.SIGWINCH)
```
Sh4d1, line 91: We should split the syscall in another `_unix.go` file and a dummy one in a `_windows.go` file with the righ build tags

internal/gotty/client.go
jerome-quere: tty.scaleway.com do not work for AMS

Sh4d1: :scream: :heavy_check_mark:

internal/gotty/client.go
```diff
@@ -0,0 +1,207 @@
...
+	// Create a chanel that will receive all resizes signals
+	resizeChan := make(chan os.Signal, 1)
+	unsubscribe := subscribeToResize(resizeChan)
+	defer unsubscribe()
+
+	wsChan, wsErrChan := websocketReader(conn)
+	cnsChan, cnsErrChan := consoleReader(cns)
+
+	// Force first resize
+	resizeChan <- syscall.SIGWINCH
```
Sh4d1, line 97: same as above, need to be in another file

PR #911, opened 2020-04-24
feat(registry): add login/logout commands

Sh4d1:
Needs #909


internal/namespaces/registry/v1/custom_namespace_login.go
```diff
@@ -0,0 +1,109 @@
...
+	secretKey, ok := client.GetSecretKey()
+	if !ok {
+		return nil, fmt.Errorf("could not get secret key")
+	}
+
+	var commandArgs []string
+
+	switch program(args.Program) {
+	case docker, podman:
+		commandArgs = []string{"login", "-u scaleway", "--password-stdin", endpoint}
```
jerome-quere, line 91:
```suggestion
		commandArgs = []string{"login", "-u", "scaleway", "--password-stdin", endpoint}
```

internal/namespaces/registry/v1/custom_namespace_login.go
```diff
@@ -0,0 +1,109 @@
...
+	default:
+		return nil, fmt.Errorf("unknown program")
+	}
+
+	loginCommand := exec.Command(args.Program, commandArgs...)
+
+	loginCommand.Stdin = strings.NewReader(secretKey)
+	loginCommand.Stderr = os.Stderr
```
jerome-quere, line 99: why not binding stdout ?

jerome-quere: We should use streams that were passed to BoostrapConfig

internal/namespaces/registry/v1/custom_namespace_login.go
```diff
@@ -0,0 +1,109 @@
...
+
+	if err := loginCommand.Run(); err != nil {
+		if execErr, ok := err.(*exec.ExitError); ok {
+			return nil, &core.CliError{Empty: true, Code: execErr.ExitCode()}
+		}
+		return nil, fmt.Errorf("could not log in to namespace: %s", err)
```
jerome-quere, line 105:
```suggestion
		return nil, fmt.Errorf("could not login to namespace: %w", err)
```

internal/namespaces/registry/v1/custom_namespace_login.go
```diff
@@ -0,0 +1,109 @@
...
+	}
+
+	return fmt.Sprintf("Successfully logged in to %s", endpoint), nil
```
jerome-quere, line 108: Use core.SuccessResult type

internal/namespaces/registry/v1/custom_namespace_login.go
Sh4d1: If we bind stoud we have the `Login Suceeded` message from docker

remyleone:
I get the following message:

```
$ go run ./cmd/scw/main.go registry  login   
{"message":"","error":{},"details":"","hint":"Valid arguments are: "}
Error saving credentials: error storing credentials - err: exit status 1, out: `{"message":"","error":{},"details":"","hint":"Valid arguments are: "}`
exit status 1
```

Could you indicate in a log message that you are using a precise executable?

```
$ go run ./cmd/scw/main.go registry  login --help
This command will run the correct command in order to log you in on the registry with the chosen program.
You will need to have the chosen binary installed on your system and in your PATH.
```

internal/namespaces/registry/v1/custom_login.go
```diff
@@ -0,0 +1,74 @@
...
+		},
+		Run: registryLoginRun,
+	}
+}
+
+func registryLoginRun(ctx context.Context, argsI interface{}) (i interface{}, e error) {
+	args := argsI.(*registryLoginArgs)
+	client := core.ExtractClient(ctx)
+
+	region := args.Region.String()
```
remyleone, line 44: Now that #1010 was merged could this code be removed?

PR #961, opened 2020-04-27
test: add tool to override exec

jerome-quere:


internal/core/exec.go
```diff
@@ -0,0 +1,28 @@
...
+		return exitErr.ExitCode(), nil
+	}
+	return 0, err
+}
+
+func ExecCmd(ctx context.Context, cmd *exec.Cmd) (exitCode int, err error) {
+	meta := extractMeta(ctx)
+
+	// We do not support override of stdin
+	cmd.Stdout = os.Stdin
```
Sh4d1, line 23: We need a way to bring the stdin in the command (for instance registry login, the stdin is a io.Reader built from the secret key)

Sh4d1: Also it's `cmd.Stdin` :smile:

internal/core/testing.go
```diff
@@ -237,6 +261,16 @@ func Test(config *TestConfig) func(t *testing.T) {
 			overideEnv["HOME"] = dir
 		}
 
+		overrideExec := defaultOverrideExec
+		if config.OverrideExec != nil {
```
Sh4d1, line 265: So we can only override 1 exec in each command? Enough for today though

internal/namespaces/instance/v1/custom_server_ssh.go
```diff
@@ -82,23 +81,24 @@ func instanceServerSSHRun(ctx context.Context, argsI interface{}) (i interface{}
...
-		"ssh", serverResp.Server.PublicIP.Address.String(),
+	sshArgs := []string{
+		serverResp.Server.PublicIP.Address.String(),
 		"-p", fmt.Sprintf("%d", args.Port),
 		"-l", args.Username,
-		"-t", args.Command,
-	)
+		"-t",
+	}
+	if args.Command != "" {
```
Sh4d1, line 90: If we split, the `-t` can be added if the command is not empty imho

PR #963, opened 2020-04-27
refactor(core): move flag parsing to bootstrap

### LINKED ISSUE #936: Use debug flag before cobra prerun
Currently, –debug flag is implemented with cobra persistent flags. As a result, the debug logger is not available before the prerun step. It could be great to directly use pflag and move the [debug logger init](https://github.com/scaleway/scaleway-cli/blob/v2/internal/core/cobra_utils.go#L118-L122) in bootstrap function.

Proposition:

- Add debug bool in BootStrapConfig. This flag is inited in main based on -D flag.
- We should keep cobra just for usage documentation (add a code comment to document that)
- Make sure the autocomplete -- notation still works.
- Add for instance a log debug message in bootstrap to see if the debug are showing up
- When the scoped logger is implemented we should be able to create a new logger in the Bootstrap function ad make sure it is accessible everywhere using context

jerome-quere:
relates to #964
Fixes #936

Blocked by https://github.com/scaleway/scaleway-sdk-go/pull/395

### DIFF
--- a/internal/core/build_info_test.go
+++ b/internal/core/build_info_test.go
@@ -14,6 +14,7 @@ import (
 var fakeCommand = &Command{
 	Namespace: "plop",
 	ArgsType:  reflect.TypeOf(args.RawArgs{}),
+	NoClient:  true,
 	Run: func(ctx context.Context, argsI interface{}) (i interface{}, e error) {
 		return &SuccessResult{}, nil
 	},
--- a/internal/core/client.go
+++ b/internal/core/client.go
@@ -174,7 +174,7 @@ func validateClient(client *scw.Client) error {
 		}
 	}
 
-	if !validation.IsRegion(defaultZone.String()) {
+	if !validation.IsRegion(defaultRegion.String()) {
 		regions := []string(nil)
 		for _, z := range scw.AllRegions {
 			regions = append(regions, string(z))
--- a/internal/core/client.go
+++ b/internal/core/client.go
@@ -93,7 +93,8 @@ More info: https://github.com/scaleway/scaleway-sdk-go/tree/master/scw#scaleway-
 	)
 }
 
-// validateProfile validate the final profile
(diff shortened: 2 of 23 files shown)


internal/core/context.go
```diff
@@ -36,9 +36,9 @@ const (
 	metaContextKey contextKey = iota
 )
 
-// newMetaContext creates a new ctx with injected meta and returns it.
-func newMetaContext(meta *meta) context.Context {
-	return context.WithValue(context.Background(), metaContextKey, meta)
+// injectMeta creates a new ctx with injected meta and returns it.
```
Sh4d1, line 39: nit: modifies the given context

jerome-quere: Context are imutable object in Go. `WithValue` create a new child context of the on given as parameter

Sh4d1: Yep but it's based on the context passed to the function

Sh4d1: Something like `creates a new ctx based on the given one in the arguments`

internal/core/client.go
```diff
@@ -96,81 +94,94 @@ More info: https://github.com/scaleway/scaleway-sdk-go/tree/master/scw#scaleway-
 }
 
 // validateProfile validate the final profile
```
Sh4d1, line 96: Same nit for the comment

internal/core/build_info_test.go
```diff
@@ -14,6 +14,7 @@ import (
 var fakeCommand = &Command{
 	Namespace: "plop",
 	ArgsType:  reflect.TypeOf(args.RawArgs{}),
+	NoClient:  true,
```
remyleone, line 17: Should we use a term such as `AllowAnonymousClient` ?

jerome-quere: tracked in #968

PR #969, opened 2020-04-28
feat: add feedback command

### LINKED ISSUE #953: Create a feedback command
Create an `scw report` or `scw feedback` command.

It opens a new issue page on the repo with a pre-filled template containing the information of the user's OS, the version of the CLI, etc ...

Rémy Leone use that for cloud code and he filled a lot of issues with that

Also, we could add a message in the scw main help text to advertise this command

remyleone:
Closes #953


internal/namespaces/feedback/custom.go
```diff
@@ -0,0 +1,91 @@
...
+			if err != nil {
+				return nil, err
+			}
+			issue := issue{
+				IssueTemplate: bug,
+				BuildInfoStr:  buildInfoStr,
+			}
+			err = issue.openInBrowser(ctx)
+			if err != nil {
+				log.Fatal(err)
```
jerome-quere, line 51: You should return the error and let the CLI print the error

internal/namespaces/feedback/custom.go
```diff
@@ -0,0 +1,91 @@
...
+			if err != nil {
+				return nil, err
+			}
+			issue := issue{
+				IssueTemplate: feature,
+				BuildInfoStr: buildInfoStr,
+			}
+			err = issue.openInBrowser(ctx)
+			if err != nil {
+				log.Fatal(err)
```
jerome-quere, line 82: Same remark as upper

internal/namespaces/feedback/issue.go
```diff
@@ -0,0 +1,121 @@
...
+type issueTemplate string
+
+const (
+	bug = issueTemplate("bug")
+	feature = issueTemplate("feature")
+)
+
+type issue struct {
+	IssueTemplate issueTemplate
+	BuildInfoStr  string
```
jerome-quere, line 27: You could use a BuildInfo object here and marshal it in `renderTemplate`

PR #998, opened 2020-05-04
feat(instance): add terminate command

### LINKED ISSUE #921: Add support for instance terminate action
It could take a very long time to shut down a GP1-L because volume needs to be backed up.

To improve this duration we could rely on instance terminate feature. We need to be careful as this will delete some volume and IP also automatically (we need to check instance API)

We would create in `instance/v1/custom_server.go`

```
func serverTerminateCommand() *core.Command
```

We would also need to add test usage in `custom_server_test.go` with the usage of this command.

In the command usage, we must be explicit if block volumes are deleted by the API when using terminate

jawher:
This command uses the server action API: with `action=terminate` https://developers.scaleway.com/en/products/instance/api/#post-049a5b to quickly delete an instance without backing up its attached volumes.

Options:

* `with-ip`: also delete the flexible IPs
* `with-block`: by default, `terminate` will delete the block storage volumes. When `false`, the command will detach the block volumes before calling `terminate`

## What remains to be decided:

* default value for `with-block`: `true` to match the API/SDK or `false` to avoid accidental data loss (though when you call `terminate`, you should expect stuff to be ... terminated.

Closes #921 

```release-note
TODO
```


remyleone:
Maybe we could add an interactive question when with-block is not specified.

jawher:
@remyleone terminate will now prompt the user iff:

* neither `with-block=true` nor `with-block=false` is specified
* the instance has at least one block volume attached

example:

```
$ scw instance server terminate 9b8ca1c9-f062-4d58-8a62-b6f8c7d0fff7 with-ip=true
Do you want to also delete the block volumes attached to this instance ? (y/N):
successfully detached volume cli-srv-goofy-rubin-1
successfully deleted ip x.x.x.x
✅ Success.
```

internal/namespaces/instance/v1/custom_server.go
```diff
@@ -822,3 +826,165 @@ func serverDeleteCommand() *core.Command {
...
+	case withBlockFalse:
+		return true, nil
+	case withBlockPrompt:
+		for _, volume := range server.Server.Volumes {
+			if volume.VolumeType != instance.VolumeTypeBSSD {
+				continue
+			}
+
+			deleteVolumes, err := interactive.PromptBoolWithConfig(&interactive.PromptBoolConfig{
+				Prompt:       "Do you want to also delete the block volumes attached to this instance ?",
```
jerome-quere, line 981:
```suggestion
				Prompt:       "Do you also want to delete block volumes attached to this instance ?",
```

internal/namespaces/instance/v1/custom_server.go
```diff
@@ -822,3 +826,165 @@ func serverDeleteCommand() *core.Command {
...
+}
+
+func shouldPreserveBlockVolumes(server *instance.GetServerResponse, terminateWithBlock withBlock) (bool, error) {
+	switch terminateWithBlock {
+	case withBlockTrue:
+		return false, nil
+	case withBlockFalse:
+		return true, nil
+	case withBlockPrompt:
+		for _, volume := range server.Server.Volumes {
```
jerome-quere, line 975:
```suggestion
		// Only prompt user if at least one block volume is attached to the instance
		for _, volume := range server.Server.Volumes {
```

internal/namespaces/instance/v1/custom_server.go
```diff
@@ -822,3 +826,165 @@ func serverDeleteCommand() *core.Command {
...
+					_, _ = interactive.Printf("successfully deleted ip %s\n", server.Server.PublicIP.Address.String())
+				}
+			}
+
+			return &core.SuccessResult{}, multiErr
+		},
+	}
+}
+
+func shouldPreserveBlockVolumes(server *instance.GetServerResponse, terminateWithBlock withBlock) (bool, error) {
```
jerome-quere, line 968: IMHO it could be easier to read if `shouldPreserveBlockVolumes` would be coded as `shouldDeleteBlockVolumes` as this would avoid insertion of the `with-block` flag

internal/namespaces/instance/v1/custom_server_test.go
```diff
@@ -279,6 +279,49 @@ func Test_ServerDelete(t *testing.T) {
...
+	interactive.IsInteractive = true
+
+	t.Run("without IPs", core.Test(&core.TestConfig{
+		Commands:   GetCommands(),
+		BeforeFunc: core.ExecStoreBeforeCmd("Server", "scw instance server create image=ubuntu-bionic -w"),
+		Cmd:        `scw instance server terminate {{ .Server.ID }}`,
+		Check: core.TestCheckCombine(
+			core.TestCheckGolden(),
+			core.TestCheckExitCode(0),
+		),
```
jerome-quere, line 294: We should add a CheckFunc that test that IP still exist by doing a `GetIP` on the API

internal/namespaces/instance/v1/custom_server_test.go
```diff
@@ -279,6 +279,49 @@ func Test_ServerDelete(t *testing.T) {
...
+	}))
+
+	t.Run("with IPs", core.Test(&core.TestConfig{
+		Commands:   GetCommands(),
+		BeforeFunc: core.ExecStoreBeforeCmd("Server", "scw instance server create image=ubuntu-bionic -w"),
+		Cmd:        `scw instance server terminate {{ .Server.ID }} with-ip=true`,
+		Check: core.TestCheckCombine(
+			core.TestCheckGolden(),
+			core.TestCheckExitCode(0),
+		),
```
jerome-quere, line 306: We should add a CheckFunc that test that IP GetIP on the API return a not found

internal/namespaces/instance/v1/custom_server_test.go
```diff
@@ -279,6 +279,49 @@ func Test_ServerDelete(t *testing.T) {
...
+		DisableParallel: true,
+	}))
+
+	t.Run("without block", core.Test(&core.TestConfig{
+		Commands:   GetCommands(),
+		BeforeFunc: core.ExecStoreBeforeCmd("Server", "scw instance server create image=ubuntu-bionic additional-volumes.0=block:10G -w"),
+		Cmd:        `scw instance server terminate {{ .Server.ID }} with-ip=true with-block=false`,
+		Check: core.TestCheckCombine(
+			core.TestCheckGolden(),
+			core.TestCheckExitCode(0),
```
jerome-quere, line 316: We should add CheckFunc that test the volume still exists

internal/namespaces/instance/v1/custom_server.go
jawher: ok for `Delete` instead of `Preserve`.

No sure I understood this part:

> as this would avoid insertion of the with-block flag

internal/namespaces/instance/v1/custom_server_test.go
jawher: I just tried, and the problem is: `GetIP` just after a `DeleteIP` returns a 403 instead of 404.

Not sure if this is intentional/stable and should be relied on in tests ?

internal/namespaces/instance/v1/custom_server_test.go
jawher: I added another test with `with-block=true`, then added a `Check` to ensure that the volume is deleted.

Problem is: the volume deletion is not immediate (after an instance is terminated), and so the check fails because the volume is still there.

jawher: cf. https://github.com/scaleway/scaleway-cli/issues/1000

internal/namespaces/instance/v1/custom_server.go
```diff
@@ -949,11 +949,13 @@ func serverTerminateCommand() *core.Command {
 
 			var multiErr error
 			if terminateServerArgs.WithIP && server.Server.PublicIP != nil && !server.Server.PublicIP.Dynamic {
+				fmt.Printf("delete IP...\n")
```
jerome-quere, line 952: Can you use `interactive.Print` instead of `fmt.Printf`. Using printf like this would break command like `scw xxxx | jq`

internal/namespaces/instance/v1/custom_server.go
```diff
@@ -822,3 +826,167 @@ func serverDeleteCommand() *core.Command {
...
+				Zone:     terminateServerArgs.Zone,
+				ServerID: terminateServerArgs.ServerID,
+				Action:   instance.ServerActionTerminate,
+			}); err != nil {
+				return nil, err
+			}
+
+			var multiErr error
+			if terminateServerArgs.WithIP && server.Server.PublicIP != nil && !server.Server.PublicIP.Dynamic {
+				fmt.Printf("delete IP...\n")
```
remyleone, line 952: @jerome-quere could we use a logger here?

internal/namespaces/instance/v1/custom_server.go
```diff
@@ -822,3 +826,167 @@ func serverDeleteCommand() *core.Command {
...
+				err = api.DeleteIP(&instance.DeleteIPRequest{
+					Zone: terminateServerArgs.Zone,
+					IP:   server.Server.PublicIP.ID,
+				})
+				if err != nil {
+					fmt.Printf("failed to delete iop: %+v\n", err)
```
remyleone, line 958: @jerome-quere Could we use a logger here?

internal/namespaces/instance/v1/custom_server.go
jawher: Whoops, that was a debug statement which was accidentally committed 🤦 

I'll delete it, to match the behaviour of the delete command

PR #1003, opened 2020-05-06
feat(qa): add ignore error filtering

remyleone:


internal/qa/ignore.go
```diff
@@ -0,0 +1,140 @@
...
+}
+
+func areCommandsEquals(c1 *core.Command, c2 *core.Command) bool {
+	return (c1.Namespace == c2.Namespace) &&
+		(c1.Verb == c2.Verb) &&
+		(c1.Resource == c2.Resource)
+}
+
+var ignoredErrors = []ignoredError{
+	{
```
jerome-quere, line 48: This is a problem. Just because uninstall do not require a zone we should not ignore all other commands.
Maybe we should not apply the DifferentLocalizationForNamespaceError for commands that explicitly set AnnonymousClient to true ?

PR #1008, opened 2020-05-12
feat(core): add a context scoped logger

remyleone:
### Community Note

Relates OR Closes #0000

```release-note

```


remyleone:
We should use this to log everything that is happening in beforefunc and afterfunc execution

internal/core/logger_test.go
```diff
@@ -0,0 +1,49 @@
...
+			},
+		},
+	)
+}
+
+func Test_Logger(t *testing.T) {
+	t.Run("Simple", Test(&TestConfig{
+		Commands: testLogCommands(),
+		BeforeFunc: func(ctx *BeforeFuncCtx) error {
+			ctx.Logger.level = logger.LogLevelDebug
```
jerome-quere, line 29: Log are not include in golden file so this do not test anything

remyleone: Could we change the `ctx.Logger.writer` to an `io.Writer` that writes on a file. Then we could check with an assert function that this file matches what the logger should have written? Like a golden log file?

PR #1036, opened 2020-05-28
fix: use correct sentry DSN

remyleone:


jerome-quere:
Does it work ?

remyleone:
@jerome-quere indeed the URL work tested it from the command line.

PR #1109, opened 2020-06-18
chore(testing): add more logging to the test utils functions

remyleone:
### Community Note

Relates OR Closes #0000

```release-note

```


internal/core/testing.go
```diff
@@ -329,7 +329,7 @@ func Test(config *TestConfig) func(t *testing.T) {
 		executeCmd := func(args []string) interface{} {
 			stdoutBuffer := &bytes.Buffer{}
 			stderrBuffer := &bytes.Buffer{}
-			logger.Debugf("command: %s", args)
+			log.Debugf("command: %s", args)
```
jerome-quere, line 332:
```suggestion
			log.Debugf("executing command: %s", args)
```

internal/core/testing.go
```diff
@@ -370,7 +372,7 @@ func Test(config *TestConfig) func(t *testing.T) {
 		if len(args) > 0 {
 			stdout := &bytes.Buffer{}
 			stderr := &bytes.Buffer{}
-			logger.Debugf("command: %s", args)
+			log.Debugf("command: %s\n", args)
```
jerome-quere, line 375:
```suggestion
			log.Debugf("executing command: %s\n", args)
```

internal/core/testing.go
```diff
@@ -386,6 +388,7 @@ func Test(config *TestConfig) func(t *testing.T) {
 				Logger:           log,
 			})
 
+			log.Debugf("Store in meta (with key CmdResult): %s\n", result)
```
jerome-quere, line 391:
```suggestion
			log.Debugf("Command %s give result: %s\n", args, result)
```

jerome-quere: The result will almost never be readable because it's a complex GoStruct. Not sure If we shoulod print it. Maybe json_encode it ?

internal/core/testing.go
jerome-quere: Wondering of all the `executing command`  should be put in the bootstrap function 🤔

remyleone: It would make sense for me to see only command, before and afterfunc execution. Why are execute commands required here?

Issue #1169, opened 2020-07-03
add support for generating markdown documentation from the CLI


remyleone:
## Description

Some users want to navigate the example from a web page. We could generate one like this on Github.
This page would be made from the different documentation strings (examples, see also) that would be put together in a markdown file that would be added to the repository in the docs/ folder.

## Version

version     2.0.0-beta.4&#43;dev
build-date  unknown
go-version  go1.14.4
git-branch  unknown
git-commit  unknown
go-arch     amd64
go-os       darwin

jerome-quere:
We should not limit ourself to the example but also generate documentation for commands

PR #1181, opened 2020-07-06
doc: add documentation generator

jerome-quere:
### Community Note

Relates to #1169


internal/docgen/tpl.go
```diff
@@ -0,0 +1,69 @@
+package docgen
+
+const tplStr = `
+
+
+{{- define "namespace" -}}
+{{ template "do_not_edit" }}
+# Documentation for scw {{ .Cmd.Namespace }}
```
remyleone, line 8:
```suggestion
# Documentation for `scw {{ .Cmd.Namespace }}`
```

remyleone: Looks like a bad idea as we cannot escape simple backtick in golang in raw litterals :(

internal/docgen/docgen.go
```diff
@@ -0,0 +1,164 @@
...
+}
+
+func renderNamespace(data *tplNamespace) (string, error) {
+	buffer := bytes.Buffer{}
+	err := newTemplate().ExecuteTemplate(&buffer, "namespace", data)
+	if err != nil {
+		return "", err
+	}
+	str := buffer.String()
+	//str = interactive.UnIndent(str, 2)
```
jerome-quere, line 98: remove

internal/docgen/docgen.go
```diff
@@ -0,0 +1,164 @@
...
+	return str, nil
+}
+
+func newTemplate() *template.Template {
+	tpl := template.New("index")
+	tpl = tpl.Funcs(map[string]interface{}{
+		"unindent": func(indent int, str string) string {
```
jerome-quere, line 105: Don't think this is used

internal/docgen/docgen.go
```diff
@@ -0,0 +1,164 @@
...
+			return "```"
+		},
+		"map": func(args ...interface{}) map[string]interface{} {
+			res := map[string]interface{}{}
+			for i := 0; i < len(args); i += 2 {
+				res[args[i].(string)] = args[i+1]
+			}
+			return res
+		},
+		"join": func(sep string, slice []string) string {
```
jerome-quere, line 121: Is this used ?

internal/docgen/docgen.go
```diff
@@ -0,0 +1,164 @@
...
+			return strings.Join(slice, sep)
+		},
+		"concat": func(strs ...string) string {
```
jerome-quere, line 124: Is this used ?

internal/interactive/print.go
```diff
@@ -61,6 +61,17 @@ func Indent(str string, indent int) string {
 	return strings.Join(lines, "\n")
 }
 
+func UnIndent(str string, indent int) string {
```
jerome-quere, line 64: I think this can be removed

PR #1268, opened 2020-07-20
feat(lb): add command to import an instance as a backend

remyleone:
### Community Note

Relates OR Closes #0000

```release-note

```


jerome-quere:
Should't this be like adding a server but with an instance ID instead of an IP  ?

```
scw lb backend add-server ip=XXXXX
scw lb backend add-server instance-server-id=XXXXX
scw lb backend add-server baremetal-server-id=XXXXX
```

remyleone:
add-server is confusing because it is in the backend resource I would expect as a user to add an IP/baremetal-server-IP/instance-server-ip to an existing backend because add suppose that a backend already exist to add them to. I would stick with import and ok for baremetal, ip. We can  add later on a new command for updating a backend that would be an `add-server`

PR #1346, opened 2020-08-17
Update docker instructions to match v2

ebriand:
Hey 👋 

I tried to use the cli inside a container following the documentation but it seems it doesn't match the v2.

I updated the documentation by adding an explicit tag (maybe too specific? But it the "v2" doesn't seem to exist). I also changed the volume since it doesn't work if you init the configuration within docker 🐳 

WDYT?


remyleone (approved):

remyleone:
Thanks a lot for your contribution @ebriand :)

PR #1380, opened 2020-09-03
feat(init): save project_id in config

QuentinBrosse:
### Community Note

```release-note
Save the `project_id` in the config with the `scw init` command
```


jerome-quere (approved):
LGTM

remyleone (approved):

jerome-quere:
Do we have test for init ? Should we add a test for this ?

QuentinBrosse:
> Do we have test for init ? Should we add a test for this?

@jerome-quere in this precise use-case it's complex to test as we need to test the behavior where the `scw init` will automatically get the `access_key`, `organization_id`, `project_id`, ... from the account API. It will require to moke the account API but without cassettes to get a consistent result and test this result.
I think that's why [we only check the `secret_key` in the "Override" test](https://github.com/scaleway/scaleway-cli/blob/master/internal/namespaces/init/init_test.go#L144).

jerome-quere (approved):
LGTM

PR #1384, opened 2020-09-03
fix(lb): improve error message when hint and details are empty

remyleone:
### Community Note

Relates OR Closes #0000

```release-note

```

### DIFF
--- a/internal/core/error.go
+++ b/internal/core/error.go
@@ -6,6 +6,7 @@ import (
 	"strings"
 
 	"github.com/scaleway/scaleway-cli/internal/human"
+	"github.com/scaleway/scaleway-sdk-go/scw"
 )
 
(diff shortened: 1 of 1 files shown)


internal/core/error.go
```diff
@@ -68,6 +69,11 @@ func (s *CliError) MarshalHuman() (string, error) {
 		sections = append(sections, str)
 	}
 
+	if sdkResponseError, isResponseError := s.Err.(*scw.ResponseError); isResponseError && s.Details == "" && s.Hint == "" {
```
jerome-quere, line 72: Are SDK error always wrapped in a CliError 🤔 ? If we explicitly catch an SDK error to put it inside a cli error this means we know what we were doing. Maybe this should be handled in the printer in case an SDK error bubble all the way up to the printer.

jerome-quere (changes requested):

PR #1453, opened 2020-09-21
feat: Adding support of ed25519 ssh key during scw init

jeansebastienh:
### Community Note

### Details
Fix the error during `scw init`:  "Except for SSH key: could not find an SSH key at ~/.ssh/id_rsa.pub" if you have a key in another format.

```release-note
NONE
```

### DIFF
--- a/internal/namespaces/account/v2alpha1/custom.go
+++ b/internal/namespaces/account/v2alpha1/custom.go
@@ -42,7 +42,7 @@ func InitRun(ctx context.Context, argsI interface{}) (i interface{}, e error) {
 	var shortenedFilename string
 	var err error
 	var localSSHKeyContent []byte
-	for _, keyName := range [3]string{"id_dsa.pub", "id_ed25519.pub", "id_rsa.pub"}  {
+	for _, keyName := range [3]string{"id_ecdsa.pub", "id_ed25519.pub", "id_rsa.pub"}  {
 		// element is the element from someSlice for where we are
 		relativePath := path.Join(".ssh", keyName)
 		filename := path.Join(core.ExtractUserHomeDir(ctx), relativePath)
--- a/internal/namespaces/account/v2alpha1/custom.go
+++ b/internal/namespaces/account/v2alpha1/custom.go
@@ -39,18 +39,23 @@ func initCommand() *core.Command {
 
 func InitRun(ctx context.Context, argsI interface{}) (i interface{}, e error) {
 	// Get default local SSH key
-	relativePath := path.Join(".ssh", "id_rsa.pub")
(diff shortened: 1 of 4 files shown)


internal/namespaces/account/v2alpha1/custom.go
```diff
@@ -39,18 +39,23 @@ func initCommand() *core.Command {
...
 	// Get default local SSH key
-	relativePath := path.Join(".ssh", "id_rsa.pub")
-	filename := path.Join(core.ExtractUserHomeDir(ctx), relativePath)
-	shortenedFilename := "~/" + relativePath
-	localSSHKeyContent, err := ioutil.ReadFile(filename)
-
+	var shortenedFilename string
+	var err error
+	var localSSHKeyContent []byte
+	for _, keyName := range [3]string{"id_dsa.pub", "id_ed25519.pub", "id_rsa.pub"}  {
```
Sh4d1, line 45: Do we really want to use DSA? :scream:

Sh4d1: Another approach could be listing `*.pub` is the `.ssh` directory, and asking the user to chose one?

jeansebastienh: Good point.
ssh supports the following formats: rsa, dsa, ecdsa and ed25519.

Which one do we want to support ?

Sh4d1: IMHO DSA should not be supported since it's been proven insecure, and is deprecated.

Also, I personally do not rely on `id_xxx` and prefer the name my keys, but we could just imply that a lot of users have 1 basic key :thinking: so this solution is still okay for me :smile:

jeansebastienh: I replaced dsa with ecdsa https://github.com/scaleway/scaleway-cli/pull/1453/commits/acd4db90f4b519dce64091baf2097c1598f97042

jerome-quere (approved):

internal/namespaces/init/custom_init_ssh_test.go
```diff
@@ -14,10 +14,14 @@ import (
 )
 
 func setUpSSHKeyLocally(key string) core.BeforeFunc {
```
jerome-quere, line 16: IMHO we should remove this function that is only called 2 times and only keep `setUpSSHKeyLocallyWithKeyName`

jerome-quere (changes requested):

internal/namespaces/init/custom_init_ssh_test.go
jeansebastienh: :+1:

remyleone (approved):

jerome-quere (approved):

PR #1525, opened 2020-11-12
feat(instance): import file for cloud-init

jtherin:
Add support of file import for cloud-init
$ scw instance server create image=ubuntu_focal name=test cloud-init=@/tmp/test


Sh4d1 (dismissed):
LGTM

Sh4d1 (dismissed):
LGTM

jtherin:
Reimplemented using ArgSpecs.

jtherin:
Hmm, seems not working on user-data, will investigate tomorrow.

internal/core/arg_file_content.go
```diff
@@ -0,0 +1,42 @@
...
+		}
+
+		fieldName := strcase.ToPublicGoName(argSpec.Name)
+		fieldValues, err := getValuesForFieldByName(reflect.ValueOf(cmdArgs), strings.Split(fieldName, "."))
+		if err != nil {
+			continue
+		}
+
+		for _, v := range fieldValues {
+			if v.Kind() != reflect.String {
```
jerome-quere, line 26: What about `StringPointer` ? is it handled in `getValuesForFieldByName` ?

jerome-quere (changes requested):
We should also:
- Add a least on test in the core package
- Add documentation for flag that have `CanLoadFile: true`
- Handle Autocompletion (maybe in another MR , if so we need an issue)

QuentinBrosse:
> Add documentation for flags that have CanLoadFile: true

@jtherin can you please add at least an example too 🙏

internal/core/arg_file_content.go
```diff
@@ -0,0 +1,56 @@
...
+func loadArgsFileContent(cmd *Command, cmdArgs interface{}) error {
+	for _, argSpec := range cmd.ArgSpecs {
+		if !argSpec.CanLoadFile {
+			continue
+		}
+
+		fieldName := strcase.ToPublicGoName(argSpec.Name)
+		fieldValues, err := getValuesForFieldByName(reflect.ValueOf(cmdArgs), strings.Split(fieldName, "."))
+		if err != nil {
+			continue
```
jerome-quere, line 24: Maybe we should return an error and not fail silently ?

jerome-quere (changes requested):
see coment

internal/core/arg_file_content.go
jtherin: validate.go line 48 et 78 also execute getValuesForFieldByName, log an info and continue.
It would not be too much to log the same info many times ?

jerome-quere (approved):

PR #1954, opened 2021-08-25
fix: allow _ in validArgName

abarbare:
This PR allow the usage of `_` in argument name. 

This is a problem for example on RDB as it's not possible to create databases or users with `_` via cli even if the RDB API allows it 

```
scw rdb database create scw_test
Invalid argument 'scw_test': arg name must only contain lowercase letters, numbers or dashes
```


abarbare:
My bad, with cli it's mandatory to specify the arg_name

i.e: `scw rdb database create name=scw_test instance-id=<uuid>` 

closing this PR

Issue #1431, opened 2020-09-14
Have a boolean flag to reveal whether a backup is exported or not


remyleone:
## Description

In `scw rdb backup list` I have no way to know whether a given backup is ready to be downloaded or not. Could we have a boolean that shows if an exportedURL is available or not.

## Version

Version    2.0.0&#43;dev
BuildDate  unknown
GoVersion  go1.15.2
GitBranch  unknown
GitCommit  unknown
GoArch     amd64
GoOS       darwin

PR #2484, opened 2022-09-01
feat(rdb): add a boolean to reveal if a backup is exported

yfodil:
feat #1431

### DIFF
--- a/internal/namespaces/rdb/v1/custom_backup.go
+++ b/internal/namespaces/rdb/v1/custom_backup.go
@@ -119,6 +119,68 @@ func backupRestoreBuilder(c *core.Command) *core.Command {
 	return c
 }
 
+func backupListBuilder(c *core.Command) *core.Command {
+	type customBackup struct {
+		ID         string                   `json:"ID"`
+		Name       string                   `json:"name"`
+		InstanceID string                   `json:"instance_ID"`
+		Exported   bool                     `json:"exported"`
+		Status     rdb.DatabaseBackupStatus `json:"status"`
+	}
+
+	c.View = &core.View{
+		Fields: []*core.ViewField{
+			{
+				Label:     "ID",
+				FieldName: "ID",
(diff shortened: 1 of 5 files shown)


internal/namespaces/rdb/v1/custom_backup_test.go
```diff
@@ -111,3 +111,36 @@ func Test_DownloadBackup(t *testing.T) {
...
 }
+
+func Test_ListBackup(t *testing.T) {
+	t.Run("Simple", core.Test(&core.TestConfig{
+		Commands: GetCommands(),
+		BeforeFunc: core.BeforeFuncCombine(
+			createInstance(engine),
+			core.BeforeFuncWhenUpdatingCassette(
+				func(ctx *core.BeforeFuncCtx) error {
+					time.Sleep(1 * time.Minute)
```
remyleone, line 122: Why do we have an hard coded sleep here?

Monitob (approved):

internal/namespaces/rdb/v1/custom_backup.go
```diff
@@ -119,6 +119,68 @@ func backupRestoreBuilder(c *core.Command) *core.Command {
...
+				expired = true
+			} else {
+				expired = time.Now().After(*backup.DownloadURLExpiresAt)
+			}
+			res = append(res, customBackup{
+				ID:         backup.ID,
+				Name:       backup.Name,
+				Status:     backup.Status,
+				InstanceID: backup.InstanceID,
+				Exported:   !expired,
```
remyleone, line 175: Could you manipulate an exported variable that will get modified across the function instead of an expired variable that sounds similar but is different?

remyleone: You could also export this functionality in an isExported function

remyleone (approved):

PR #2487, opened 2022-09-02
feat(config): add destroy command

yfodil:


internal/namespaces/rdb/v1/testdata/test-list-backup-simple.cassette.yaml
```diff
@@ -914,7 +914,7 @@ interactions:
     url: https://api.scaleway.com/rdb/v1/regions/fr-par/backups/da73454a-8232-4d91-b773-8f16ad50a7e1
```
remyleone, line 914: Could you remove those files from this PR they are absolutely not related.

internal/namespaces/config/commands_test.go
```diff
@@ -210,6 +210,19 @@ func Test_ConfigDumpCommand(t *testing.T) {
...
 }
 
+func Test_ConfigDestroyCommand(t *testing.T) {
+	t.Run("Simple", core.Test(&core.TestConfig{
+		Commands:   GetCommands(),
+		BeforeFunc: beforeFuncCreateFullConfig(),
+		Cmd:        "scw config destroy",
+		Check: core.TestCheckCombine(
+			core.TestCheckExitCode(0),
+			core.TestCheckGolden(),
```
remyleone, line 220: Add a check that verifies that no configuration file is present.

internal/namespaces/config/commands_test.go
```diff
@@ -210,6 +210,51 @@ func Test_ConfigDumpCommand(t *testing.T) {
 	}))
 }
 
+func Test_ConfigDestroyCommand(t *testing.T) {
+	path := "/tmp/test_config_destroy/"
```
remyleone, line 214: Use `TmpHomeDir`

remyleone: For the TmpHomeDir @yfodil  prefer to create my own temporary directory where I can set a path that @yfodil  define, using TmpHomeDir the folders in local and on the CI will not be the same so the Golden will not be valid:

remyleone (approved):

PR #2511, opened 2022-09-15
feat(core): allow disabling version check on build

Codelax:
Allows packagers to disable version check when building cli


seblu:
Tested and working ; but I also have errors in the test suite.

Issue #2217, opened 2022-03-29
missing scw instance image update


jtherin:
It's possible to set an instance image as public, this allow any organization to use it if they know the uuid.

## Command attempted

scw instance image update $UUID zone=fr-par-3 public=true

### Expected Behavior

set image $UUID as public

### Actual Behavior

Print help message with missing update action.

```$ scw instance image update
Images are backups of your instances.
You can reuse that image to restore your data or create a series of instances with a predefined configuration.

An image is a complete backup of your server including all volumes.

USAGE:
  scw instance image <command>

AVAILABLE COMMANDS:
  create      Create an instance image
  delete      Delete an instance image
  get         Get an instance image
  list        List instance images
  wait        Wait for image to reach a stable state

FLAGS:
  -h, --help   help for image

GLOBAL FLAGS:
  -c, --config string    The path to the config file
  -D, --debug            Enable debug mode
  -o, --output string    Output format: json or human, see 'scw help output' for more info (default "human")
  -p, --profile string   The config profile to use

Use "scw instance image [command] --help" for more information about a command.
```

PR #2539, opened 2022-10-06
feat(instance): add image update command

yfodil:
feat #2217

### DIFF
--- a/internal/namespaces/instance/v1/custom_image.go
+++ b/internal/namespaces/instance/v1/custom_image.go
@@ -420,15 +420,15 @@ func imageUpdateCommand() *core.Command {
 		Examples: []*core.Example{
 			{
 				Short: "Update image name",
-				Raw:   "scw instance image update image-id=173814d5-7fbb-463c-8650-ec511c7f28e5 name=foo",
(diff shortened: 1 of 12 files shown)


internal/namespaces/instance/v1/custom_image.go
```diff
@@ -314,6 +315,125 @@ func imageDeleteBuilder(c *core.Command) *core.Command {
 	return c
 }
 
+//
```
remyleone, line 318: Why is it not generated?

yfodil: Because the [UpdateImage](https://github.com/scaleway/scaleway-sdk-go/blob/master/api/instance/v1/image_utils.go#L81) method has been hand coded in the SDK

remyleone: Could we check with the instance team if the update image could be generated now? Maybe the reasons of why it was hardcoded are no long true and it could be generated just like any other function

remyleone: @kindermoumoute

remyleone (approved):

PR #2549, opened 2022-10-14
feat(rdb): add instance update settings

yfodil:
feat #2198


internal/namespaces/rdb/v1/custom_instance.go
```diff
@@ -163,6 +163,184 @@ func instanceUpgradeBuilder(c *core.Command) *core.Command {
...
+					return nil, err
+				}
+			}
+
+			updateInstanceResponse, err := api.UpdateInstance(updateInstanceRequest)
+			if err != nil {
+				return nil, err
+			}
+
+			resp, err := api.GetInstance(&rdb.GetInstanceRequest{
```
quantumsheep, line 308: No need to fetch the instance after updating it, `UpdateInstance` returns the updated instance.

internal/namespaces/rdb/v1/custom_instance.go
```diff
@@ -163,6 +163,184 @@ func instanceUpgradeBuilder(c *core.Command) *core.Command {
...
+					Region:     updateInstanceRequest.Region,
+					InstanceID: updateInstanceRequest.InstanceID,
+					Settings:   settings,
+				})
+				if err != nil {
+					return nil, err
+				}
+			}
+
+			updateInstanceResponse, err := api.UpdateInstance(updateInstanceRequest)
```
remyleone, line 303: The return of the update method already contains the Instance. You don't need to do a second GetInstance after.

internal/namespaces/rdb/v1/custom_instance.go
```diff
@@ -163,6 +163,184 @@ func instanceUpgradeBuilder(c *core.Command) *core.Command {
...
+				settings := getResp.Settings
+				changes := customRequest.Settings
+
+				for _, change := range changes {
+					matched := false
+					for _, setting := range settings {
+						if change.Name == setting.Name {
+							setting.Value = change.Value
+							matched = true
+							break
```
remyleone, line 285: What if 2 settings are updated? Will the second one still happen with this break?

internal/namespaces/rdb/v1/custom_instance_test.go
```diff
@@ -56,6 +58,58 @@ func Test_UpgradeInstance(t *testing.T) {
...
+			core.TestCheckGolden(),
+			core.TestCheckExitCode(0),
+		),
+		AfterFunc: deleteInstance(),
+	}))
+
+	t.Run("Modify default max_connections from 100 to 200", core.Test(&core.TestConfig{
+		Commands:   GetCommands(),
+		BeforeFunc: createInstance("PostgreSQL-12"),
+		Cmd:        "scw rdb instance update {{ .Instance.ID }} settings.0.name=max_connections settings.0.value=200 --wait",
```
remyleone, line 104: Could you try to add a test where you modify 3 settings ?

internal/namespaces/rdb/v1/custom_instance.go
yfodil: Yes

yfodil: I meant yes, the user can update existing settings and add new ones too in a single command

remyleone (approved):

PR #2550, opened 2022-10-17
fix(core): do not require a config file if required env vars are set

jawher:
## Issue

A regression was introduced in the 2.6.0 release where the CLI will exit with the following error:

```
scaleway-sdk-go: cannot read config file /root/.config/scw/config.yaml: no such file or directory
```

Even though all the required env vars `SCW_ACCESS_KEY`, `SCW_SECRET_KEY`, `SCW_ORGANIZATION_ID`, `SCW_DEFAULT_ORGANIZATION_ID`, `SCW_ZONE` are set.

The same invocation of the CLI used to work fine with the 2.5.x releases

## Proposed Fix

Change the client creation logic to no longer require a configuration file, but instead:

* first load config from env
* if a config file is found, then load it and merge it with the env config

```release-note
fix regression where a config file was required even if all required env vars are set
```

### DIFF
--- a/internal/core/client.go
+++ b/internal/core/client.go
@@ -28,7 +28,7 @@ func createClient(ctx context.Context, httpClient *http.Client, buildInfo *Build
 	config, err := scw.LoadConfigFromPath(ExtractConfigPath(ctx))
 	switch {
 	case errIsConfigFileNotFound(err):
-		// no config file wass found -> nop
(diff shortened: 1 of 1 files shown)


Codelax (approved):
LGTM
Thanks for your contribution

remyleone (approved):

PR #2575, opened 2022-11-08
feat(lb): improve backend get output

yfodil:


internal/namespaces/lb/v1/custom_backend.go
```diff
@@ -21,3 +21,48 @@ var (
...
+func lbMarshalerFunc(i interface{}, opt *human.MarshalOpt) (string, error) {
+	type tmp lb.LB
+	loadbalancer := tmp(i.(lb.LB))
+
+	opt.Sections = []*human.MarshalSection{
+		{
+			FieldName: "IP",
+		},
+		{
+			FieldName: "Instances",
```
remyleone, line 34: Not necessarily an instance. Could be a baremetal or even something completely different

remyleone: Why not Backends?

remyleone: Also why is Frontend missing?

remyleone (approved):

PR #2699, opened 2022-12-27
feat(instance): improve pnic create argument validation

yfodil:


internal/namespaces/instance/v1/custom_privatenics.go
```diff
@@ -31,3 +33,27 @@ func privateNicListBuilder(c *core.Command) *core.Command {
...
+		if err == nil {
+			return res, nil
+		}
+
+		resErr, isResErr := err.(*scw.InvalidArgumentsError)
+		if !isResErr {
+			return nil, err
+		}
+
+		if resErr.Details[0].ArgumentName == "private_network_id" && resErr.Details[0].HelpMessage == "required key not provided" {
```
remyleone, line 49: Why is this not handled by command generation? Could we mark an argument as required and let the generation handle it?

yfodil:
Will be handled on code generation side

PR #2715, opened 2023-01-06
feat(instance): get pnic from mac address

yfodil:


internal/namespaces/instance/v1/custom_privatenics.go
```diff
@@ -31,3 +32,38 @@ func privateNicListBuilder(c *core.Command) *core.Command {
...
+	c.Interceptor = func(ctx context.Context, argsI interface{}, runner core.CommandRunner) (interface{}, error) {
+		tmpRequest := argsI.(*instance.GetPrivateNICRequest)
+
+		if isMacAddress(tmpRequest.PrivateNicID) {
+			client := core.ExtractClient(ctx)
+			api := instance.NewAPI(client)
+
+			listPrivateNICs, err := api.ListPrivateNICs(&instance.ListPrivateNICsRequest{
+				Zone:     tmpRequest.Zone,
+				ServerID: tmpRequest.ServerID,
```
remyleone, line 48: Is it possible to filter here by mac address? It would avoid to do a complete loop

remyleone (approved):

PR #2757, opened 2023-02-07
fix(config): catch profile error even when env vars are set

yfodil:


internal/core/client.go
```diff
@@ -38,6 +38,19 @@ func createClient(ctx context.Context, httpClient *http.Client, buildInfo *Build
...
 		}
 
+		opts := []scw.ClientOption{
+			scw.WithDefaultRegion(scw.RegionFrPar),
+			scw.WithDefaultZone(scw.ZoneFrPar1),
+			scw.WithUserAgent(buildInfo.GetUserAgent()),
+			scw.WithProfile(activeProfile),
+			scw.WithHTTPClient(httpClient),
+		}
+
```
remyleone, line 48: Maybe add a comment here. Also, if you don't need all those options do not hesitate to remove them. You should have a check that is focused on active profile if I understand correctly

remyleone (approved):

PR #2784, opened 2023-02-17
feat(core): arg resolver

Codelax:
This is only a draft of how we could resolve arguments


Codelax:
This was achieved partially in completion with [AutocompleteGetArg](https://github.com/scaleway/scaleway-cli/blob/1e94877b5fd50aff2baf146ec1414375cdbb649e/internal/core/autocomplete_utils.go#L90)

PR #2785, opened 2023-02-20
fix(instance): remove hardcoded offers

Codelax:
DEV1-S is not available in certain AZ, we remove the default type instead of changing every time Scaleway change its hardware

### DIFF
--- a/internal/namespaces/instance/v1/custom_server_create_test.go
+++ b/internal/namespaces/instance/v1/custom_server_create_test.go
@@ -133,10 +133,10 @@ func Test_CreateServer(t *testing.T) {
 		t.Run("valid single local snapshot", core.Test(&core.TestConfig{
 			Commands: instance.GetCommands(),
 			BeforeFunc: core.BeforeFuncCombine(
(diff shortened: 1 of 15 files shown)


internal/namespaces/instance/v1/custom_server_create_test.go
```diff
@@ -497,7 +497,7 @@ func Test_CreateServerErrors(t *testing.T) {
 
 	t.Run("Error: invalid total local volumes size: too low 2", core.Test(&core.TestConfig{
 		Commands: instance.GetCommands(),
-		Cmd:      "scw instance server create image=ubuntu_jammy root-volume=l:5GB additional-volumes.0=block:10GB",
+		Cmd:      testServerCommand("image=ubuntu_bionic root-volume=l:5GB additional-volumes.0=block:10GB"),
```
Codelax, line 500:
```suggestion
		Cmd:      testServerCommand("image=ubuntu_jammy root-volume=l:5GB additional-volumes.0=block:10GB"),
```

remyleone (approved):

PR #2862, opened 2023-03-13
feat(tasks): change api

Codelax:


remyleone (approved):

internal/namespaces/config/commands.go
```diff
@@ -584,6 +599,11 @@ func configInfoCommand() *core.Command {
 				}
 			}
 
+			if len(overridedVariables) > 0 {
+				msg := "Some variables are overridden by the environment: " + strings.Join(overridedVariables, ", ")
+				fmt.Println(terminal.Style(msg, color.FgRed))
```
quantumsheep:
```suggestion
				fmt.Fprintln(os.Stderr, terminal.Style(msg, color.FgOrange))
```

Better to display it as a warning and in stderr (so that it can be ignored in case of being use with piping)

quantumsheep (changes requested):

quantumsheep (approved):

remyleone (approved):

PR #2879, opened 2023-03-20
fix(secret-manager): Move secret content in doc

Cubox:
The documentation contains a potentially harmful example, setting the secret in the description, instead of the proper data field.


remyleone:
this file is automatically generated using scw-doc-gen

remyleone:
Fixed with https://github.com/scaleway/scaleway-cli/pull/2881

PR #2963, opened 2023-03-28
feat(vpcgw): improve gateway-type list visualization

yfodil:
feat #2895


internal/namespaces/vpcgw/v1/testdata/test-list-gateway-type-simple.golden
```diff
@@ -0,0 +1,21 @@
+🎲🎲🎲 EXIT CODE: 0 🎲🎲🎲
+🟩🟩🟩 STDOUT️ 🟩🟩🟩️
+Types:
```
quantumsheep, line 3: It might be better to return only the `Types` field, given that it is the sole field returned.

quantumsheep (changes requested):

quantumsheep (approved):

PR #3018, opened 2023-04-11
feat(config): add validate command

yfodil:


internal/namespaces/config/commands.go
```diff
@@ -695,6 +696,43 @@ func configImportCommand() *core.Command {
...
+			// validate the remaining profiles
+			for _, profile := range config.Profiles {
+				err = validateProfile(profile)
+				if err != nil {
+					return nil, err
+				}
+			}
+
+			return &core.SuccessResult{
+				Message: "successfully validate config",
```
quantumsheep, line 730:
```suggestion
				Message: "configuration is valid",
```

internal/namespaces/config/commands.go
```diff
@@ -749,3 +787,67 @@ func getProfile(config *scw.Config, profileName string) (*scw.Profile, error) {
...
+
+func validateProfile(profile *scw.Profile) error {
+	if profile.AccessKey != nil {
+		if *profile.AccessKey == "" {
+			return &core.CliError{
+				Err: fmt.Errorf("access key cannot be empty"),
+			}
+		} else if !validation.IsAccessKey(*profile.AccessKey) {
+			return core.InvalidAccessKeyError(*profile.AccessKey)
+		}
```
quantumsheep, line 799:
```suggestion
		if *profile.AccessKey == "" {
			return &core.CliError{
				Err: fmt.Errorf("access key cannot be empty"),
			}
		}
		
		if !validation.IsAccessKey(*profile.AccessKey) {
			return core.InvalidAccessKeyError(*profile.AccessKey)
		}
```
No need for `else` (same for other checks)

quantumsheep (changes requested):

internal/namespaces/config/commands.go
yfodil: It follows the same pattern as other config commands, we may think to rephrase theses messages

cmd/scw/testdata/test-all-usage-config-validate-usage.golden
```diff
@@ -0,0 +1,15 @@
+🎲🎲🎲 EXIT CODE: 0 🎲🎲🎲
+🟥🟥🟥 STDERR️️ 🟥🟥🟥️
+Validate the config
```
remyleone, line 3: What does it mean? Will it test for YAML correctness? Will it check that the config contains only valid fields? Please explain more about what this validation will mean.

internal/namespaces/config/commands.go
```diff
@@ -749,3 +787,79 @@ func getProfile(config *scw.Config, profileName string) (*scw.Profile, error) {
 	}
 	return profile, nil
 }
+
+func validateProfile(profile *scw.Profile) error {
+	if profile.AccessKey != nil {
```
remyleone, line 792: Could you extract functions that perform all those operations? Maybe it could be helpful to reuse them at different place in the CLI and SDK?

remyleone (approved):

quantumsheep (changes requested):

quantumsheep (approved):

PR #3024, opened 2023-04-13
feat: wasm build

### LINKED ISSUE #2674: core: webasm build and interface
It could be integrated into a web application
Multiple things should be sorted out:
- Remove commands that are specific to a system (using os.Exec, docker)
- Possible use of a shell with an interface for completion
- An interface to run cli commands from webasm with authentication without env variables
- Test within the CI that webasm work

Codelax:
Closes #2674


wasm/package.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "scaleway-cli-wasm",
+  "version": "1.0.0",
```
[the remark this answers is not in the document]
Codelax: Not published yet, version will be set on publish by the CI. Same version as CLI would be the best

remyleone (dismissed):

internal/config/config.go
```diff
@@ -57,6 +58,13 @@ type Config struct {
 // returns a new empty config if file doesn't exist
 // return error if fail to load config file
 func LoadConfig(configPath string) (*Config, error) {
+	if runtime.GOARCH == "wasm" {
+		return &Config{
+			Alias:  alias.EmptyConfig(),
+			Output: DefaultOutput,
+			path:   configPath,
+		}, nil
+	}
```
quantumsheep, line 67:
```suggestion
	}

```

internal/core/client.go
```diff
@@ -5,15 +5,31 @@ import (
...
+	}
+
+	return scw.NewClient(opts...)
+}
+
 // createClient creates a Scaleway SDK client.
 func createClient(ctx context.Context, httpClient *http.Client, buildInfo *BuildInfo, profileName string) (*scw.Client, error) {
+	if runtime.GOARCH == "wasm" {
+		return createWasmClient(httpClient, buildInfo)
+	}
```
quantumsheep, line 32:
```suggestion
	}

```

wasm/wasm_exec.d.ts
```diff
@@ -0,0 +1,8 @@
+export class Go {
+    argv: string[]
+    env: {[key: string]: string}
```
quantumsheep, line 3:
```suggestion
    env: Record<string, string>
```

wasm/wasm_exec.d.ts
```diff
@@ -0,0 +1,8 @@
...
+    importObject: {
+        go: {[key: string]: (sp: number) => void},
```
quantumsheep, line 5:
```suggestion
        go: Record<string, (sp: number) => void>,
```

cmd/scw-wasm/main.go
```diff
@@ -0,0 +1,81 @@
...
+		if arg.Type() != js.TypeString {
+			return nil, fmt.Errorf("invalid argument at index %d", argIndex)
+		}
+		cliArgs = append(cliArgs, arg.String())
+	}
+
+	exitCodeChan := runCommand(cliArgs, stdout, stderr)
+	exitCode := <-exitCodeChan
+	if exitCode != 0 {
+		errBody := stderr.String()
```
quantumsheep, line 56: `stdout` might also be wanted here

quantumsheep (changes requested):

quantumsheep (changes requested):
Minor changes

cmd/scw-wasm/main.go
Codelax: This error happen when a cli command is executed, it should not have something on stdout when an error occurs.
If a command return `value, err` and error is not nil then the error will be printed to stderr and the value will be dismissed.
What should be improved is using stderr when a command is successful, this is where warnings will be displayed.
Both output should be merged

PR #3063, opened 2023-04-21
feat(instance): server create using snapshots

### LINKED ISSUE #1905: Command for restoring an instance using a snapshot
## Description

In the restore instance part, one solution I wanted to apply was 

1. Creating a snapshot first.
2. Power off the instance in order to detach the volume 
3. Attach the volume of the snapshot to the instance.  

## How this functionality would be exposed 

I thought I could use this command 

`scw instance server attach-volume server-id=your-server-id volume-id=your-volume-id` replacing the volume-id with the snapshot id to attach the volume. Can you make it work? 

Another option would be to have a restore verb inside the server resource.

## Version

Version    2.3.0
BuildDate  unknown
GoVersion  go1.16
GitBranch  unknown
GitCommit  unknown
GoArch     amd64
GoOS       darwin

Codelax:
Closes #1905 
Can be used as
`scw instance server create root-volume=local:<snapshot_id> additional-volumes.0=block:<snapshot_id>`

### DIFF
--- a/internal/namespaces/instance/v1/custom_server_create_test.go
+++ b/internal/namespaces/instance/v1/custom_server_create_test.go
@@ -125,6 +125,26 @@ func Test_CreateServer(t *testing.T) {
 			AfterFunc: deleteServerAfterFunc(),
 		}))
 
+		t.Run("valid single local snapshot", core.Test(&core.TestConfig{
+			Commands: GetCommands(),
+			BeforeFunc: core.BeforeFuncCombine(
+				core.ExecStoreBeforeCmd("Server", "scw instance server create image=ubuntu_bionic root-volume=local:20GB stopped=true"),
+				core.ExecStoreBeforeCmd("Snapshot", `scw instance snapshot create volume-id={{ (index .Server.Volumes "0").ID }}`),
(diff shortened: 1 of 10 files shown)


internal/namespaces/instance/v1/custom_server_create_test.go
```diff
@@ -385,6 +426,26 @@ func Test_CreateServerErrors(t *testing.T) {
...
+		Check: core.TestCheckCombine(
+			core.TestCheckGolden(),
+			core.TestCheckExitCode(1),
+		),
+		DisableParallel: true,
+	}))
+
+	t.Run("Error: invalid additional volume snapshot ID", core.Test(&core.TestConfig{
+		Commands: GetCommands(),
+		Cmd:      "scw instance server create image=ubuntu_bionic additional-volumes.0=block:29da9ad9-e759-4a56-82c8-f0607f93055c",
```
remyleone, line 441: Why is this id hard coded?

Codelax: This is a test that should output an error

PR #3092, opened 2023-04-25
ci: strip binaries and check size

### LINKED ISSUE #3042: Strip binary when doing a release
Our binaries should be as stripped as possible. We should take a look at the size of the binary as a metric we keep an eye on

Codelax:
Closes #3042

### DIFF
--- a/.github/workflows/unit-tests.yaml
+++ b/.github/workflows/unit-tests.yaml
@@ -46,6 +46,10 @@ jobs:
           fetch-depth: 1
       - name: Build binaries
         run: ./scripts/build.sh
+      - name: Print binaries size
+        run: ls -lh ./bin
+      - name: Check binary size
+        run: ./scripts/check-size.sh ./bin/*linux-x86_64
 
   docker-tests:
     runs-on: ubuntu-latest
--- a/scripts/check-size.sh
+++ b/scripts/check-size.sh
@@ -1,5 +1,7 @@
 #/bin/env bash
 
+MAX_BINARY_SIZE=25000000
+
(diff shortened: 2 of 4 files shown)


scripts/check-size.sh
```diff
@@ -0,0 +1,26 @@
...
+
+if [[ $# -ne 1 ]]; then
+  usage
+fi
+
+BINARY_SIZE=$(stat -c %s $1)
+
+echo "Binary size: $BINARY_SIZE"
+
+if (( $BINARY_SIZE > 25000000 )); then
```
remyleone, line 21: Can you make it a constant ?

.github/workflows/unit-tests.yaml
```diff
@@ -46,6 +46,10 @@ jobs:
           fetch-depth: 1
       - name: Build binaries
         run: ./scripts/build.sh
+      - name: Print binaries size
+        run: ls -lh ./bin
+      - name: Check binary size
+        run: ./scripts/check-size.sh ./bin/*linux-x86_64
```
remyleone, line 52: Why is it only linux binaries?

Codelax: Different architecture have different sizes. There is only on configured limit that I chose based on linux_amd64 build.
For example, darwin_arm already is bigger than this limit, we should keep only one as reference.
This check does not prevent us from increasing the binary size but you will need to increase the limit in the culprit PR so it will be clear when adding a feature that the CLI increase in size.

remyleone (approved):

PR #3119, opened 2023-05-11
feat(wasm): add jwt support

### LINKED ISSUE #3112: wasm: jwt support
Allow the use of jwt when running command using js functions.

This should be passed as a configuration object in the `run` function, ex: 
```ts
const run = async (cfg: RunOptions, ...args []string)
type RunOptions = {
  JWT: string
}
```

Codelax:
Fix #3112 
Added a config passed to the js `run` function.
Some platform dependent code was moved from `core` to `platform`. Will be improved when adding output configuration (termcaps)
Added a `jshelpers` package to handle js values

Had to fix some tests as by moving client creation it showed that ReloadClient skipped the client validation


internal/core/context.go
```diff
@@ -26,14 +27,14 @@ type meta struct {
 	OverrideEnv  map[string]string
 	OverrideExec OverrideExecFunc
 	CliConfig    *cliConfig.Config
+	Platform     platform.Platform
```
remyleone, line 30: Can/Should we change the user agent if the CLI is doing a request in the webassembly mode ? How can we track usage for it?

remyleone (approved):

PR #3122, opened 2023-05-12
[doc] add missing param on upgrade pool command example

bmm-alc:
The parameter region is mandatory on upgrade pool command

### Community Note

Relates OR Closes #0000

```release-note
NONE
```

### DIFF
--- a/docs/commands/k8s.md
+++ b/docs/commands/k8s.md
@@ -1012,7 +1012,7 @@ scw k8s pool upgrade <pool-id ...> [arg=value ...]
 
 Upgrade a specific pool to the Kubernetes version 1.24.7
 ```
-scw k8s pool upgrade 11111111-1111-1111-111111111111 version=1.24.7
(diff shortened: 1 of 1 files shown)


Codelax:
Hi
As stated in the command usage, the `region` argument is optional.
```
[region=fr-par]   Region to target. If none is passed will use default region from the config (fr-par | nl-ams | pl-waw)
```

The CLI handle locality arguments (zone, region) by using a default (fr-par-1, fr-par) or using the default one in config.
The config values can be configured using:
- `scw config set default-region=...`
- Editing config file in `~/.config/scw/config.yml`
- Using environment variables (`SCW_DEFAULT_REGION`)

More info can be found in the `scw config` [usage](https://github.com/scaleway/scaleway-cli/blob/master/docs/commands/config.md).

If you had issue with region argument being missing this might be a bug. As executing the command with bare minimal config should just use `fr-par`.
If so don't hesitate to open an issue.
Thanks for your interest with the CLI.

PR #3148, opened 2023-05-26
feat(wasm): export an autocomplete function

### LINKED ISSUE #3111: wasm: autocomplete function
Export functions to interact with CLI autocomplete. If possible the AutoCompleteTree should be generated only one time.

### LINKED ISSUE #3114: wasm: correct typings
The types in the wasm package should be improved an cleaned up.

An `index.d.ts` can be created at the package root, it can be added to package.json as `"types": "index.d.ts"`.

This type index should export all types, the ones for wasm_exec, the CLI and the current index.js

Codelax:
Closes #3111 
Closes #3114 
Currently not using directly autocomplete as it requires a filled Context, will be done in the future.

The completion tree is built upon each call.

### DIFF
--- a/internal/wasm/autocomplete.go
+++ b/internal/wasm/autocomplete.go
@@ -1,0 +1,40 @@
+package wasm
+
+import (
+	"fmt"
+	"strconv"
+	"strings"
+)
+
+type AutoCompleteConfig struct {
+	JWT          string   `js:"jwt"`
+	LeftWords    []string `js:"leftWords"`
+	SelectedWord string   `js:"selectedWord"`
+	RightWords   []string `js:"rightWords"`
+}
+
+func Autocomplete(cfg *AutoCompleteConfig) ([]string, error) {
+	indexToComplete := int64(len(cfg.LeftWords) + 2)
+
+	words := append(cfg.LeftWords, cfg.SelectedWord)
+	words = append(words, cfg.RightWords...)
+
+	completeCommand := []string{"autocomplete", "complete", "zsh", strconv.FormatInt(indexToComplete, 10), "scw"}
+
+	completeCommand = append(completeCommand, words...)
+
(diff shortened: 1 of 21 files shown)


internal/wasm/autocomplete.go
```diff
@@ -0,0 +1,40 @@
...
+	RightWords   []string `js:"rightWords"`
+}
+
+func Autocomplete(cfg *AutoCompleteConfig) ([]string, error) {
+	indexToComplete := int64(len(cfg.LeftWords) + 2)
+
+	words := append(cfg.LeftWords, cfg.SelectedWord)
+	words = append(words, cfg.RightWords...)
+
+	completeCommand := []string{"autocomplete", "complete", "zsh", strconv.FormatInt(indexToComplete, 10), "scw"}
```
remyleone, line 22: Could ZSH be in the AutoCompleteConfig? Maybe depending on the different feature of a shell we want to configure them differently?

wasm/src/jshelpers.test.ts
```diff
@@ -0,0 +1,40 @@
+
```
remyleone, line 1: Is there a linter running on all the typescript code that we got?

Codelax: Not currently

internal/wasm/autocomplete.go
remyleone: Could you add a comment about this directly in the code to give directions about how to refactor this code if you think it is going to be complex and time-consuming?

wasm/src/jshelpers.test.ts
remyleone: https://github.com/scaleway/scaleway-cli/issues/3154

remyleone (approved):

PR #3152, opened 2023-05-30
feat(k8s): create PN automatically with cluster if none given

Mia-Cross:
This PR adds a custom `clusterCreate` function that creates a Private Network with the cluster if no `private_network_id` is given.


jtherin:
Beware, that will break kosmos!

remyleone (approved):

internal/namespaces/k8s/v1/custom_cluster.go
```diff
@@ -104,6 +108,82 @@ func clusterCreateBuilder(c *core.Command) *core.Command {
 		return runner(ctx, args)
 	}
 
+	c.Run = func(ctx context.Context, args interface{}) (i interface{}, e error) {
```
remyleone, line 111: need to handle kosmos case

remyleone (changes requested):

internal/namespaces/k8s/v1/custom_cluster.go
```diff
@@ -104,6 +106,84 @@ func clusterCreateBuilder(c *core.Command) *core.Command {
...
+
+		client := core.ExtractClient(ctx)
+		k8sAPI := k8s.NewAPI(client)
+		vpcAPI := vpc.NewAPI(client)
+
+		pnCreated := false
+		var pn *vpc.PrivateNetwork
+		var err error
+
+		if request.Type != "multicloud" {
```
jtherin, line 120: il n'y a pas que multicloud comme offer kosmos, plutot ce qui commence par kapsule

remyleone: strings.Prefix de kapsule ?

PR #3279, opened 2023-07-12
feat(vpc): only support regional vpc

yfodil:


internal/namespaces/vpc/v1/custom_private_network.go
```diff
@@ -1,326 +0,0 @@
...
-	"github.com/scaleway/scaleway-sdk-go/api/instance/v1"
-	"github.com/scaleway/scaleway-sdk-go/api/lb/v1"
-	"github.com/scaleway/scaleway-sdk-go/api/rdb/v1"
-	"github.com/scaleway/scaleway-sdk-go/api/redis/v1"
-	"github.com/scaleway/scaleway-sdk-go/api/vpc/v1"
-	"github.com/scaleway/scaleway-sdk-go/api/vpcgw/v1"
-	"github.com/scaleway/scaleway-sdk-go/scw"
-)
-
-func privateNetworkGetBuilder(c *core.Command) *core.Command {
```
remyleone, line 17: Is this function and set of features migrated to v2 as well?

yfodil: For a matter of time, I'll do it later in another PR

PR #3297, opened 2023-07-21
feat(instance): routed ip support

Codelax:
### Community Note

Relates OR Closes #0000

```release-note

```

### DIFF
--- a/internal/human/marshal_func.go
+++ b/internal/human/marshal_func.go
@@ -80,7 +80,12 @@ func init() {
 	})
 	marshalerFuncs.Store(reflect.TypeOf(scw.IPNet{}), func(i interface{}, opt *MarshalOpt) (string, error) {
 		v := i.(scw.IPNet)
-		return v.String(), nil
+		str := v.String()
+		if str == "<nil>" {
+			return "-", nil
+		}
+
+		return str, nil
 	})
 	marshalerFuncs.Store(reflect.TypeOf(version.Version{}), func(i interface{}, opt *MarshalOpt) (string, error) {
 		v := i.(version.Version)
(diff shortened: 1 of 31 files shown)


internal/human/marshal_func.go
```diff
@@ -80,7 +80,12 @@ func init() {
 	})
 	marshalerFuncs.Store(reflect.TypeOf(scw.IPNet{}), func(i interface{}, opt *MarshalOpt) (string, error) {
 		v := i.(scw.IPNet)
-		return v.String(), nil
+		str := v.String()
+		if str == "<nil>" {
+			return "-", nil
+		}
```
quantumsheep, line 86: Is there a better way to handle nil `IPNet`?

quantumsheep (approved):

internal/human/marshal_func.go
Codelax: I've done it as in sdk-go: https://github.com/scaleway/scaleway-sdk-go/blob/b0dfa1defaafac939d792a76cf6ad1f28b1b0cc0/scw/custom_types.go#L216

The issue I have is that we use `scw.IPNet`. The type is:
```go
type IPNet struct {
	net.IPNet
}
```
It is itself not a pointer. If we look at how the `<nil>` is generated in Go libary we can find:
```go
func (n *IPNet) String() string {
	if n == nil {
		return "<nil>"
	}
```
I don't get it how we can get `<nil>` using a non-pointer struct :/

Codelax: Just because of this in `net.IPNet.String()`
```go
	nn, m := networkNumberAndMask(n)
	if nn == nil || m == nil {
		return "<nil>"
	}
```

This should be fine to check `<nil>`

PR #3722, opened 2024-03-21
feat(core): set can handle all positional args together

Mia-Cross:
This PR introduces a new behavior for the `set` verb when there are several positional arguments.

This feature will be helpful for #3597 but for now, there is no `set` command that is using positional arguments, so the test in this PR only shows that the already existing commands with `set` still work. More tests will be added in #3597.


internal/core/cobra_utils_test.go
```diff
@@ -247,3 +271,32 @@ func Test_PositionalArg(t *testing.T) {
...
+			core.TestCheckExitCode(0),
+			core.TestCheckGolden(),
+		),
+	}))
+
+	t.Run("set verb with multi positional", core.Test(&core.TestConfig{
+		Commands: testGetCommands(),
+		Cmd:      "scw test multi-positional pos1 pos2 pos3 tag=tag1",
+		Check: core.TestCheckCombine(
+			core.TestCheckExitCode(0),
```
Codelax, line 289: Can you check the result, it is expected that it contain list of arguments.
Last time I saw the error because of the golden but it could be missed in a larger PR.

internal/core/command.go
```diff
@@ -47,6 +47,9 @@ type Command struct {
 	// ArgSpecs defines specifications for arguments.
 	ArgSpecs ArgSpecs
 
+	// AcceptMultiplePositionalArgs defines whether the command can accept multiple positional arguments.
+	AcceptMultiplePositionalArgs bool
```
Codelax, line 51:
```suggestion
	// AcceptMultiplePositionalArgs defines whether the command can accept multiple positional arguments.
	// If enabled, positional argument is expected to be a list.
	AcceptMultiplePositionalArgs bool
```

PR #3829, opened 2024-05-23
fix(autocomplete): add basename arg for custom path

### LINKED ISSUE #1959: Add support for autocomplete behind shim (like asdf)
## Description
When using scw behind a version managed using shims like ASDF, the autocomplete doesn't work as it is generating the full path in the generated autocompletion code (even the scw shim is in PATH)

```
$ scw autocomplete script
autoload -U compinit && compinit
_/home/mpoussin/.asdf/installs/scaleway-cli/2.3.1/bin/scw () {
        output=($(/home/mpoussin/.asdf/installs/scaleway-cli/2.3.1/bin/scw autocomplete complete zsh -- ${CURRENT} ${words}))
        opts=('-S' ' ')
        if [[ $output == *= ]]; then
                opts=('-S' '')
        fi
        compadd "${opts[@]}" -- "${output[@]}"
}
compdef _/home/mpoussin/.asdf/installs/scaleway-cli/2.3.1/bin/scw /home/mpoussin/.asdf/installs/scaleway-cli/2.3.1/bin/scw
```
It would be interesting to have a `--parameter` to force the basename to `scw` (or configurable)

## How this functionality would be exposed
A new parameter `--basename` for example to force a basename on the autocompletion generated code

## References

https://github.com/asdf-vm/asdf
https://github.com/webofmars/asdf-plugin-scaleway-cli

## Version

Version    2.3.1
BuildDate  2021-06-08T11:54:02
GoVersion  go1.16.4
GitBranch  new-release
GitCommit  07dbaebc
GoArch     amd64
GoOS       linux

### LINKED ISSUE #3676: Autocompletion script wrongly generated (at least) for bash if using path to call `scw`
https://github.com/scaleway/scaleway-cli/blob/d17343311bb0f7645dea2d01ecb2cb00adbacbac/internal/namespaces/autocomplete/autocomplete.go#L81

…because the basename of `binaryName` should be extracted. In the following test, `scw` is called with full path, resulting in wrongly named bash function:

```bash
$ ~/bin/scw autocomplete script shell=bash
_/home/mid/bin/scw() {
	_get_comp_words_by_ref -n = cword words

	output=$(/home/mid/bin/scw autocomplete complete bash -- "$COMP_LINE" "$cword" "${words[@]}")
	COMPREPLY=($output)
	# apply compopt option and ignore failure for older bash versions
	[[ $COMPREPLY == *= ]] && compopt -o nospace 2> /dev/null || true
	return
}
complete -F _/home/mid/bin/scw /home/mid/bin/scw
```

:wink:

Laure-di:
close #3676 
close #1959

### DIFF
--- a/internal/namespaces/autocomplete/autocomplete.go
+++ b/internal/namespaces/autocomplete/autocomplete.go
@@ -52,7 +52,6 @@ type autocompleteScript struct {
 // autocompleteScripts regroups the autocomplete scripts for the different shells
 // The key is the path of the shell.
 func autocompleteScripts(ctx context.Context, basename string) map[string]autocompleteScript {
-	binaryName := basename
 	homePath := core.ExtractUserHomeDir(ctx)
 	return map[string]autocompleteScript{
 		"bash": {
@@ -78,8 +77,8 @@ func autocompleteScripts(ctx context.Context, basename string) map[string]autoco
 				return
 			}
 			complete -F _%[1]s %[1]s
-		`, binaryName),
-			CompleteScript: fmt.Sprintf(`eval "$(%s autocomplete script shell=bash)"`, binaryName),
+		`, basename),
+			CompleteScript: fmt.Sprintf(`eval "$(%s autocomplete script shell=bash)"`, basename),
 			ShellConfigurationFile: map[string]string{
 				"darwin": path.Join(homePath, ".bash_profile"),
 				"linux":  path.Join(homePath, ".bashrc"),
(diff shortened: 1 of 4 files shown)


Codelax (approved):

internal/namespaces/autocomplete/autocomplete.go
```diff
@@ -51,8 +51,8 @@ type autocompleteScript struct {
 
 // autocompleteScripts regroups the autocomplete scripts for the different shells
 // The key is the path of the shell.
-func autocompleteScripts(ctx context.Context) map[string]autocompleteScript {
-	binaryName := core.ExtractBinaryName(ctx)
+func autocompleteScripts(ctx context.Context, basename string) map[string]autocompleteScript {
+	binaryName := basename
```
Codelax, line 55: why not use `binaryName` as function argument ?

Laure-di: good question, because the arg is call basename but I can change it to binaryName

Codelax (approved):

PR #3863, opened 2024-06-07
feat(instance): improve support of windows based servers

Codelax:
- Add `--wait` to `scw instance server get-rdp-password`.
- Add `admin-password-encryption-ssh-key-id` with completion to `scw instance server {create,update}`
- Make `admin-password-encryption-ssh-key-id` required when creating a Windows server. This avoid confusion as password will not be generated without a ssh key id.


internal/namespaces/instance/v1/custom_server_create.go
```diff
@@ -190,20 +198,31 @@ func instanceServerCreateRun(ctx context.Context, argsI interface{}) (i interfac
...
 	}
 
 	client := core.ExtractClient(ctx)
 	apiMarketplace := marketplace.NewAPI(client)
 	apiInstance := instance.NewAPI(client)
 
+	if commercialTypeIsWindowsServer(serverReq.CommercialType) && serverReq.AdminPasswordEncryptionSSHKeyID == nil {
+		return nil, &core.CliError{
+			Err:     core.MissingRequiredArgumentError("admin-password-encryption-ssh-key-id").Err,
+			Details: "Expected a SSH Key ID to encrypt Admin RDP password. If not provided, not password will be generated. Key must be RSA Public Key.",
```
remyleone, line 219:
```suggestion
			Details: "Expected a SSH Key ID to encrypt Admin RDP password. If not provided, no password will be generated. Key must be RSA Public Key.",
```

internal/namespaces/instance/v1/custom_server_rdp_test.go
```diff
@@ -0,0 +1,143 @@
...
+// tryLoadKey will try to load an RSA SSH Key from given path.
+// If not found, it will generate the key and create the file.
+func loadRSASSHKey(path string) (*rsa.PrivateKey, error) {
+	pemContent, err := os.ReadFile(path)
+	if err != nil && !os.IsNotExist(err) {
+		return nil, fmt.Errorf("failed to read test key: %w", err)
+	}
+
+	if *core.UpdateCassettes {
+		privateKey, err := rsa.GenerateKey(rand.Reader, 2048)
```
remyleone, line 30: Is there a strong constraint on the size of the key? If so we should mention it in the documentation and have it as a constant to change it easily later on in case we need to increase the size of the key

Codelax: There is no constraint. This is a random value and my goal was to reduce file size.

remyleone (approved):

PR #3916, opened 2024-06-27
fix(dedibox): set default value project-id

Laure-di:
### Community Note

Relates OR Closes #0000

```release-note

```

### DIFF
--- a/internal/core/context.go
+++ b/internal/core/context.go
@@ -89,10 +89,7 @@ func GetOrganizationIDFromContext(ctx context.Context) string {
 
 func GetProjectIDFromContext(ctx context.Context) string {
 	client := ExtractClient(ctx)
-	projectID, exists := client.GetDefaultProjectID()
-	if !exists {
-		panic("no default project ID found")
-	}
+	projectID, _ := client.GetDefaultProjectID()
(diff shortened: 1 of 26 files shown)


internal/core/context.go
```diff
@@ -89,10 +89,7 @@ func GetOrganizationIDFromContext(ctx context.Context) string {
 
 func GetProjectIDFromContext(ctx context.Context) string {
 	client := ExtractClient(ctx)
-	projectID, exists := client.GetDefaultProjectID()
```
remyleone, line 92: Hum I'm not really comfortable with silencing this test of existence :/

Codelax: This is the same behavior as `GetOrganizationIDFromContext(ctx context.Context) string`, a couple lines before this one.
Goal is that the Default function should not fail if there is no default project id. As Default function is used when generating documentation

Laure-di: The end goal is a fix on the API side, though it's a temporary solution. This function is only used here, and if there is no project ID, the API will throw an error.

remyleone (dismissed):

remyleone (approved):

PR #4137, opened 2024-09-18
feat(k8s): rework kubeconfig handling

jtherin:


internal/namespaces/k8s/v1/custom_kubeconfig_get_test.go
```diff
@@ -3,31 +3,82 @@ package k8s_test
...
+				// config, err := yaml.Marshal(ctx.Meta["Kubeconfig"].(api.Config))
+				// assert.Equal(t, err, nil)
+				// assert.Equal(t, ctx.Result.(string), string(config))
+			},
+			core.TestCheckExitCode(0),
+		),
+		AfterFunc: deleteCluster("Cluster"),
+	}))
+
+	// t.Run("copy-token", core.Test(&core.TestConfig{
```
remyleone, line 63: Why is this code commented?

jtherin: to avoid possible leaking of a token

PR #4140, opened 2024-09-20
feat(instance): server create with custom iops volumes

Codelax:


internal/namespaces/instance/v1/custom_server_create_builder.go
```diff
@@ -314,3 +355,220 @@ func (sb *ServerBuilder) Validate() error {
...
+		if err != nil {
+			return nil, fmt.Errorf("invalid volume iops %s in %s volume", parts[2], flagV)
+		}
+		vb.IOPS = scw.Uint32Ptr(uint32(iops))
+		parts = parts[0:2]
+	}
+
+	if len(parts) == 2 {
+		switch parts[0] {
+		case "l", "local":
```
remyleone, line 424: Could you add documentation for all types of volumes, it seems complicated to find the right meaning without examples and knowing what are the different options available for a given server.

internal/namespaces/instance/v1/custom_server_create_builder.go
```diff
@@ -314,3 +355,220 @@ func (sb *ServerBuilder) Validate() error {
...
+			vb.VolumeType = instance.VolumeVolumeTypeLSSD
+		case "b", "block":
+			vb.VolumeType = instance.VolumeVolumeTypeBSSD
+		case "s", "scratch":
```
remyleone, line 428: scratch is missing from the examples

remyleone (approved):

PR #4171, opened 2024-10-07
feat(instance): change default server type from DEV1-S to PLAY2-PICO

Codelax:
DEV1-S is not available in all AZ and is subject to depreciation. Old price is €0.0137/hour and new price is €0.014/hour. Biggest difference is the lack of local storage.


remyleone:
I would be in favor of removing default choice to avoid surprising customers with unanticipated extra cost due to an upgrade in the CLI. That way the command will break in case argument is not specified and the customer will be able to make an explicit choice.

PR #4633, opened 2025-04-02
feat(instance): render block volumes details when getting a server

### LINKED ISSUE #4399: Improve visualisation of block volumes in the GetServer command
```
$ scw instance server get 18bf502f-cf10-4426-b0f6-08289edfbf61
ID                         18bf502f-cf10-4426-b0f6-08289edfbf61
...

Volumes:
ID                                    NAME  EXPORT URI  ORGANIZATION  SIZE  VOLUME TYPE  CREATION DATE  MODIFICATION DATE  STATE  PROJECT  BOOT   ZONE
39315473-a8ad-4520-a38d-e69ba644e614  -     -           -             -     sbs_volume   -              -                  -      -        false  fr-par-1
a0ab81a2-fc91-4369-a859-6fbdf6cf67a2  -     -           -             -     sbs_volume   -              -                  -      -        false  fr-par-1

Private NICs:
ID   MAC ADDRESS  PRIVATE NETWORK NAME  PRIVATE NETWORK ID
```

It would be helpful to have a better ordering of the columns that appear here (export URI does not seem to be very useful, having organization is not too useful either,
- Size is not filled
- no info on the IOPS
- No info on the state
- project is probably not very useful here

Mia-Cross:
Closes #4399

### DIFF
--- a/internal/namespaces/instance/v1/custom_server.go
+++ b/internal/namespaces/instance/v1/custom_server.go
@@ -120,15 +120,28 @@ func serversMarshalerFunc(i interface{}, opt *human.MarshalOpt) (string, error) 
 	return human.Marshal(humanServers, opt)
 }
 
+type customVolume struct {
+	ID               string     `json:"id"`
+	Name             string     `json:"name"`
+	Size             scw.Size   `json:"size"`
+	VolumeType       string     `json:"volume_type"`
+	IOPS             string     `json:"iops"`
+	State            string     `json:"state"`
+	CreationDate     *time.Time `json:"creation_date"`
+	ModificationDate *time.Time `json:"modification_date"`
+	Boot             bool       `json:"boot"`
+	Zone             string     `json:"zone"`
+}
+
 // orderVolumes return an ordered slice based on the volume map key "0", "1", "2",...
-func orderVolumes(v map[string]*instance.VolumeServer) []*instance.VolumeServer {
+func orderVolumes(v map[string]*customVolume) []*customVolume {
 	indexes := []string(nil)
 	for index := range v {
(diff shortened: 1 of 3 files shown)


internal/namespaces/instance/v1/testdata/test-create-server-errors-error-invalid-image-label.golden
```diff
@@ -1,11 +1,10 @@
 🎲🎲🎲 EXIT CODE: 1 🎲🎲🎲
 🟥🟥🟥 STDERR️️ 🟥🟥🟥️
-Cannot find resource 'MarketplaceImage' with ID 'macos'
+Internal Server Error
```
remyleone, line 3: That's suspect. Could you dig on that?

internal/namespaces/instance/v1/custom_server.go
```diff
@@ -407,13 +420,58 @@ func serverGetBuilder(c *core.Command) *core.Command {
...
+				Zone:     volume.Zone,
+			})
+			if blockVol != nil {
+				customVol.Name = blockVol.Name
+				customVol.Size = blockVol.Size
+				customVol.VolumeType = blockVol.Type
+				customVol.State = blockVol.Status.String()
+				customVol.CreationDate = blockVol.CreatedAt
+				customVol.ModificationDate = blockVol.UpdatedAt
+				if *blockVol.Specs.PerfIops == 5000 {
```
remyleone, line 444: I would add a test to ensure Specs is not nil and I would transform it in a switch case to handle more easily different kind of block volume iops

remyleone (approved):

PR #4672, opened 2025-04-16
feat(instance): support legacy b_ssd in server creation

Codelax:


internal/namespaces/instance/v1/custom_server_create_builder.go
```diff
@@ -565,6 +565,8 @@ func NewVolumeBuilder(zone scw.Zone, flagV string) (*VolumeBuilder, error) {
 			vb.VolumeType = instance.VolumeVolumeTypeLSSD
 		case "sbs", "b", "block":
 			vb.VolumeType = instance.VolumeVolumeTypeSbsVolume
+		case "b_ssd":
```
pierre-muller, line 568: Do we want to add a feature which will be deprecated in the coming weeks (April 30th)?

PR #4766, opened 2025-05-27
fix(instance): display sbs volumes' size in `instance image list` details

Mia-Cross:
The command `scw instance image list` now fetches volume information from the SBS API to display the correct size of the underlying snapshots.


internal/namespaces/instance/v1/custom_image.go
```diff
@@ -210,6 +211,33 @@ func imageListBuilder(c *core.Command) *core.Command {
 			newCustomImage := &imageListItem{
 				Image: image,
 			}
+
+			if image.RootVolume.VolumeType == instance.VolumeVolumeTypeSbsSnapshot {
```
remyleone, line 215: Could you add a test for this feature?

PR #4909, opened 2025-07-22
tests(all): bump go-vcr to v3.2.0

Mia-Cross:
`go-vcr` introduced a very helpful `SkipRequestLatency` option on the recorder in version 3. This will allow us to replay cassettes a lot faster because interactions will not simulate the requests' duration anymore.

This is a breaking change however, and we should re-record all cassettes with the new version.

The following changes had to be made in the tests:
- core/build_info_test.go:Test_CheckVersion
    - update expected CLI version
- core/checks_test.go:TestCheckAPIKey
    - replace API key in request URL, otherwise cassette matching fails
- internal/namespaces/container/v1beta1/custom_deploy_test.go:Test_Deploy
    - go-vcr.v3 has changes with URL encoding which break this test, so we had to implement a custom matcher for docker requests
- internal/namespaces/registry/v1/custom_docker_helper_test.go:TestRegistryDockerHelperGetCommand
    - replace Secret in golden to avoid exposing SCW_SECRET_KEY


remyleone (dismissed):

elwinar:
Stale, and the migration to v4 is already started and partially done, let's finish it instead.

PR #5636, opened 2026-06-04
feat(object): add support for 's3-endpoint' to bucket cmds

### LINKED ISSUE #5568: Add support for overriding endpoints for object storage
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

It would be helpful to have support for custom endpoints for object storage.

pypaut:
### Community Note

Closes #5568.

This PR depends on https://github.com/scaleway/scaleway-sdk-go/pull/3100.

```release-note
Add an `s3-endpoint` flag to the bucket commands
Add support for `s3_endpoint` specified in the configuration profile
```

### DIFF
--- a/cmd/scw/testdata/test-all-usage-mongodb-instance-create-usage.cassette.yaml
+++ b/cmd/scw/testdata/test-all-usage-mongodb-instance-create-usage.cassette.yaml
@@ -2,29 +2,27 @@
 version: 1
 interactions:
 - request:
-    body: '{"versions":[{"version":"7.0.12", "end_of_life_at":"2026-01-02T15:04:05Z",
-      "available_settings":[]}], "total_count":1}'
+    body: '{"versions":[{"version":"7.0.34","end_of_life_at":"2027-08-31T23:59:59Z","available_settings":[]},{"version":"7.0.30","end_of_life_at":"2027-08-31T23:59:59Z","available_settings":[]},{"version":"7.0.28","end_of_life_at":"2027-08-31T23:59:59Z","available_settings":[]},{"version":"7.0.21","end_of_life_at":"2027-08-31T23:59:59Z","available_settings":[]},{"version":"7.0.12","end_of_life_at":"2027-08-31T23:59:59Z","available_settings":[]}],"total_count":5}'
     form: {}
     headers:
       User-Agent:
-      - scaleway-sdk-go/v1.0.0-beta.33.0.20250516160029-3f13aad4f58a (go1.24.2; darwin;
-        amd64) cli-e2e-test
+      - scaleway-sdk-go/v1.0.0-beta.36.0.20260602154418-ab7d81da5600 (go1.26.4; darwin;
+        arm64) cli-e2e-test
     url: https://api.scaleway.com/mongodb/v1alpha1/regions/fr-par/versions
     method: GET
   response:
-    body: '{"versions":[{"version":"7.0.12", "end_of_life_at":"2026-01-02T15:04:05Z",
-      "available_settings":[]}], "total_count":1}'
+    body: '{"versions":[{"version":"7.0.34","end_of_life_at":"2027-08-31T23:59:59Z","available_settings":[]},{"version":"7.0.30","end_of_life_at":"2027-08-31T23:59:59Z","available_settings":[]},{"version":"7.0.28","end_of_life_at":"2027-08-31T23:59:59Z","available_settings":[]},{"version":"7.0.21","end_of_life_at":"2027-08-31T23:59:59Z","available_settings":[]},{"version":"7.0.12","end_of_life_at":"2027-08-31T23:59:59Z","available_settings":[]}],"total_count":5}'
     headers:
       Content-Length:
-      - "118"
+      - "455"
       Content-Security-Policy:
       - default-src 'none'; frame-ancestors 'none'
       Content-Type:
       - application/json
       Date:
-      - Mon, 19 May 2025 22:36:36 GMT
+      - Thu, 04 Jun 2026 13:34:13 GMT
       Server:
-      - Scaleway API Gateway (fr-par-3;edge01)
+      - Scaleway API Gateway (fr-par-1;edge01)
       Strict-Transport-Security:
       - max-age=63072000
       X-Content-Type-Options:
@@ -32,7 +30,7 @@ interactions:
       X-Frame-Options:
       - DENY
       X-Request-Id:
-      - aaaacbf5-1f08-4d9c-8a96-c7519a551691
+      - 1b636820-bb28-416a-bd39-406058135dbe
     status: 200 OK
     code: 200
     duration: ""
--- a/cmd/scw/testdata/test-all-usage-mongodb-instance-create-usage.golden
+++ b/cmd/scw/testdata/test-all-usage-mongodb-instance-create-usage.golden
@@ -8,7 +8,7 @@ USAGE:
 ARGS:
   [project-id]                                             Project ID to use. If none is passed the default project ID will be used
   [name=<generated>]                                       Name of the Database Instance
-  version=7.0.12                                           Version of the MongoDB® engine
+  version=7.0.34                                           Version of the MongoDB® engine
   [tags.{index}]                                           Tags to apply to the Database Instance
   node-number=1                                            Number of node to use for the Database Instance
   node-type                                                Type of node to use for the Database Instance
--- a/internal/namespaces/object/v1/custom_bucket.go
+++ b/internal/namespaces/object/v1/custom_bucket.go
@@ -10,7 +10,6 @@ import (
 	"reflect"
 	"strings"
 
-	"github.com/aws/aws-sdk-go-v2/config"
 	"github.com/aws/aws-sdk-go-v2/service/s3"
 	"github.com/aws/aws-sdk-go-v2/service/s3/types"
 	"github.com/scaleway/scaleway-cli/v2/core"
@@ -457,7 +456,7 @@ func getAPIEndpoint(ctx context.Context, region string, customEndpoint ...string
 	}
 
 	// AWS configuration, by environment variable or config file
-	ep, err := getAPIEndpointFromAWSConf(ctx)
+	ep, err := scw.GetS3EndpointFromAWSConf(ctx)
 	if err != nil {
 		return "", fmt.Errorf("could not get API endpoint from AWS conf: %w", err)
 	}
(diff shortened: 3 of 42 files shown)


pypaut:
Doesn't build because it relies on an SDK version that is not live yet.

pypaut:
Broken because depends on https://github.com/scaleway/scaleway-sdk-go/pull/3253

cmd/scw/testdata/test-all-usage-mongodb-instance-create-usage.cassette.yaml
remyleone, line 1: why this file is modified?

cmd/scw/testdata/test-all-usage-mongodb-instance-create-usage.golden
remyleone, line 1: Why is this file modified?

internal/namespaces/object/v1/custom_bucket_cors.go
```diff
@@ -21,6 +21,8 @@ type bucketCorsCreateArgs struct {
 	CorsConfiguration string
 	Region            scw.Region
 	ProjectID         string
+	S3Endpoint        string `json:"s3-endpoint"`
```
remyleone: Could you include a test to test those flags?

internal/namespaces/object/v1/custom_bucket.go
```diff
@@ -118,9 +132,11 @@ func bucketCreateCommand() *core.Command {
...
 
 type bucketDeleteArgs struct {
-	Region    scw.Region
-	Name      string
-	ProjectID string
+	Region         scw.Region
+	Name           string
+	ProjectID      string
+	S3Endpoint     string `json:"s3-endpoint"`
+	S3UsePathStyle bool   `json:"s3-use-path-style"`
```
remyleone: why does it use a bool while all the others are using string for `S3UsePathStyle`?

internal/namespaces/object/v1/s3configfile.go
```diff
@@ -32,12 +32,13 @@ func (s *supportedTool) ToStringArray() []string {
...
-	Name      string
-	ProjectID string
-	HomeDir   string
+	AccessKey  string
+	SecretKey  string
+	Region     scw.Region
+	Name       string
+	ProjectID  string
+	HomeDir    string
+	S3Endpoint string
```
remyleone: When is this field used?

internal/namespaces/object/v1/s3_helpers.go
```diff
@@ -154,3 +156,170 @@ func normalizeOwnerID(id *string) string {
...
+		return true
+	}
+
+	// Profile field value
+	scwClient := core.ExtractClient(ctx)
+
+	return scwClient.GetS3UsePathStyle()
+}
+
+func getBucketEndpoint(
```
remyleone: env var/profile checks unreachable since s3Endpoint is always pre-resolved; also uses hardcoded "SCW_S3_ENDPOINT" string instead of scw.ScwS3EndpointEnv constant.

internal/namespaces/object/v1/custom_bucket.go
```diff
@@ -63,16 +63,25 @@ func bucketCreateCommand() *core.Command {
...
+				Required:   false,
+				Short:      "Custom S3 endpoint to use instead of the default",
+			},
+			{
+				Name:       "s3-use-path-style",
+				Positional: false,
+				Required:   false,
+				Short:      "Whether to use path style addressing for S3 API calls or not",
 			},
+			core.ProjectIDArgSpec(),
```
estellesoulard, line 77: Did the previous description mean we were expected to input the project-id with iam access key syntax (`api_key@project_id`) ? If so the `core.ProjectIDArgSpec()` has a generic description, wouldn't it be better to keep the previous custom description ?

internal/namespaces/object/v1/custom_bucket.go
estellesoulard: Other similar cmd structs in this PR like `bucketConfigArgs` have a `S3UsePathStyle string`

internal/namespaces/object/v1/custom_bucket.go
```diff
@@ -247,16 +289,19 @@ func bucketListCommand() *core.Command {
 		ArgsType:  reflect.TypeOf(bucketListArgs{}),
 		ArgSpecs: core.ArgSpecs{
```
estellesoulard, line 321: Isn't `S3UsePathStyle` missing from the ArgSpecs even though it's in the `bucketListArgs`?

internal/namespaces/object/v1/custom_bucket.go
```diff
@@ -247,16 +289,19 @@ func bucketListCommand() *core.Command {
...
 			},
+			core.ProjectIDArgSpec(),
 			core.RegionArgSpec(),
 		},
 		Run: func(ctx context.Context, argsI any) (any, error) {
 			args := argsI.(*bucketListArgs)
-			client := newS3Client(ctx, args.Region, args.ProjectID)
+
+			s3Endpoint := getS3Endpoint(ctx, args.Region.String(), args.S3Endpoint)
+			client := newS3Client(ctx, args.Region, args.ProjectID, s3Endpoint, false)
```
estellesoulard, line 343: Is the hardcoded `false` deliberate or should it be `args.S3UsePathStyle`?

internal/namespaces/object/v1/custom_bucket.go
```diff
@@ -555,7 +531,9 @@ func autocompleteBucketName(
 	}
 
 	suggestions := core.AutocompleteSuggestions(nil)
-	client := newS3Client(ctx, region, projectID)
+	s3Endpoint := getS3Endpoint(ctx, region.String(), "")
+	s3UsePathStyle := getS3UsePathStyle(ctx, "false")
```
remyleone: asses "false" instead of "", short-circuiting env var/profile fallback.

internal/namespaces/object/v1/custom_bucket.go
```diff
@@ -306,16 +351,20 @@ func bucketUpdateCommand() *core.Command {
...
 			{
-				Name:         "project-id",
-				Short:        "Scaleway project ID to use with IAM Access Key syntax",
-				Required:     false,
-				ValidateFunc: core.ValidateProjectID(),
+				Name:       "s3-endpoint",
+				Positional: false,
+				Required:   false,
+				Short:      "Custom S3 endpoint to use instead of the default",
 			},
```
remyleone, line 397: s3-use-path-style missing from ArgSpecs (flag won't be recognized).

internal/namespaces/object/v1/custom_bucket.go
remyleone, line 343: hardcodes false for path-style; env var/profile ignored.

internal/namespaces/object/v1/custom_bucket.go
```diff
@@ -140,16 +156,25 @@ func bucketDeleteCommand() *core.Command {
...
+			},
+			core.ProjectIDArgSpec(),
 			core.RegionArgSpec(),
 		},
 		Run: func(ctx context.Context, argsI any) (any, error) {
 			args := argsI.(*bucketDeleteArgs)
-			client := newS3Client(ctx, args.Region, args.ProjectID)
+
+			s3Endpoint := getS3Endpoint(ctx, args.Region.String(), args.S3Endpoint)
+			client := newS3Client(ctx, args.Region, args.ProjectID, s3Endpoint, args.S3UsePathStyle)
```
remyleone: S3UsePathStyle is bool, passed directly without getS3UsePathStyle, so env var/profile are ignored.

remyleone (changes requested):

estellesoulard:
If it's realistic, could some tests be added on each impacted call using the new flags to confirm the new behaviour?

internal/namespaces/object/v1/custom_bucket.go
pypaut: No, the format of the project ID is still expected to be the single standard UUID. I actually don't know what the previous description meant, and replaced it without noticing the difference.

Unless it had any particular meaning, I will keep it as is.

internal/namespaces/object/v1/custom_bucket.go
pypaut: It is deliberate, `S3UsePathStyle` does not make sense for the `ListBuckets` call.

internal/namespaces/object/v1/s3configfile.go
pypaut: It was not. I will fix the configuration templates with it.

internal/namespaces/object/v1/custom_bucket.go
estellesoulard: I was referring to [this syntax](https://www.scaleway.com/en/docs/iam/api-cli/using-api-key-object-storage/#overriding-the-preferred-project-when-making-a-call)

pypaut: Yes, your previous comment was clear about this 👌🏻.

The previous definition of the `project-id` argument used `ValidateFunc: core.ValidateProjectID()` which doesn't actually allow this format. The description was erroneous, and didn't really make sense since it is the access key that can tolerate the `@PROJECT_ID` suffix, not the project ID that tolerates the `ACCESS_KEY@` prefix.

Still fixed the support for such key format in [this commit](https://github.com/scaleway/scaleway-cli/pull/5636/changes/8329d06415186fbb2a28cba31bc9e47ae49d35e9).

internal/namespaces/object/v1/s3_helpers.go
```diff
@@ -154,3 +166,156 @@ func normalizeOwnerID(id *string) string {
...
+		scwClient := core.ExtractClient(ctx)
+		profileS3Endpoint, s3EndpointOk := scwClient.GetS3Endpoint()
+
+		if s3EndpointOk && profileS3Endpoint != "" {
+			// Don't use the default S3 endpoint with the default region,
+			// in case a different region was passed to this function
+
+			// Possible edge case: a user who explicitly sets s3_endpoint:
+			// https://s3.fr-par.scw.cloud in profile and runs --region nl-ams
+			// will silently get nl-ams endpoint instead of their configured one
```
estellesoulard, line 263: This is the case for the `profileS3Endpoint` , but not for an S3Endpoint set as an env var? Should this behaviour be uniform regardless of how the s3endpoint is set?

pypaut: The difference between the profile endpoint and the others (CLI arg, env var...) is that this one is automatically populated by the SDK when not set by the user. This is why we need to check whether the endpoint is exactly what the SDK would have produced or not:

```
if auto-generated (or equal to auto-generated):
    ignore this value and build the endpoint with the passed region
else:
    use this value
```

pypaut: Okay I think I misunderstood your comment. This edge case does exists in each `if` clause, and it's actually not an edge case. It should simply be documented.

docs/commands/dedibox.md
remyleone, line 1: Why is this removed?

docs/commands/index.md
```diff
@@ -1,144 +0,0 @@
-<!-- This file was created manually and acts as the landing page for the CLI documentation website. Removing it will break the documentation build. -->
```
remyleone, line 1: Why is this removed?

remyleone (changes requested):

PR #5950, opened 2026-08-04
feat(instance): add support for templates

Mia-Cross:
This PR adds support for the Instance Templates feature.

### DIFF
--- a/cmd/scw/testdata/test-all-usage-instance-usage.golden
+++ b/cmd/scw/testdata/test-all-usage-instance-usage.golden
@@ -15,6 +15,7 @@ AVAILABLE COMMANDS:
   server-type     Instance type management commands
   snapshot        Snapshot management commands
   ssh             SSH Utilities
+  template        Template management commands
   user-data       User data management commands
   volume          Volume management commands
   volume-type     Volume type management commands
--- a/internal/namespaces/instance/v2alpha1/custom.go
+++ b/internal/namespaces/instance/v2alpha1/custom.go
@@ -1,0 +1,58 @@
+package instance
+
+import (
+	"github.com/scaleway/scaleway-cli/v2/core"
+	"github.com/scaleway/scaleway-cli/v2/core/human"
+	instanceSDK "github.com/scaleway/scaleway-sdk-go/api/instance/v2alpha1"
+)
+
+// GetCommands returns instance commands.
+//
+// This function:
+// - Gets the generated commands
+// - Register handwritten marshalers
+// - Apply handwritten overrides (of Command.Run and Command.View)
+// - Merge handwritten commands
+func GetCommands() *core.Commands {
+	cmds := core.NewCommands(
+		instanceTemplate(),
+		instanceTemplateList(),
+		instanceTemplateCreate(),
+		instanceTemplateGet(),
+		instanceTemplateUpdate(),
+		instanceTemplateDelete(),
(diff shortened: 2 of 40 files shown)


cmd/scw/testdata/test-all-usage-instance-usage.golden
```diff
@@ -15,6 +15,7 @@ AVAILABLE COMMANDS:
   server-type     Instance type management commands
   snapshot        Snapshot management commands
   ssh             SSH Utilities
+  template        List templates
```
remyleone: Is it normal that is not a generic term and focus on list?

internal/namespaces/instance/v2alpha1/custom.go
```diff
@@ -0,0 +1,58 @@
...
+		instanceTemplateCheck(),
+		instanceTemplateCreateServer(),
+	)
+
+	//
+	// Templates
+	//
+	human.RegisterMarshalerFunc(instanceSDK.Template{}, templateMarshalerFunc)
+
+	cmds.MustFind("instance", "template", "create").Override(TemplateCreateBuilder)
```
pypaut, line 39: Why is this necessary here, and not for other commands e.g. `bucket`? I think I don't understand what this does.

Mia-Cross: The `cmds` object is initialized with the generated commands (found in `instance_cli.go`), but some of them might not be very easy to use, so they need customization.

The first part of the line (MustFind) ensures that `cmds` does indeed contain a command for `instance template create`, and selects it. The second part (Override) sets the TemplateCreateBuilder function as the function that adds custom logic to the selected generated command.

In this case, I added a generator for `name`, I marked the `server-type` attribute as Required and renamed the IP arguments from `public-ip-v4-count` to `public-ipv4-count` to ensure consistency with every other command of the CLI.

This logic is not needed in the `object` namespace as there are no generated commands to override, only custom functions.

pypaut: Thank you for these details. I understand that the generated commands are from the generation tools from our internal Protobuf repositories, which explains why this logic doesn't exist in `object`.

However I noticed something else: in the function `GetCommands()` (in the `instance` namespace), the description indicates the first step being `Gets the generated commands`. And while other namespaces (like `webhosting`) do make a call to `GetGeneratedCommands()`, I don't see that in `instance`. Instead, new commands relative to templates are directly declared with `core.NewCommands()`. And these new commands (relative to templates) are indeed not present in the generated commands of `instance_cli.go`.

Is there any reasons for this?

Mia-Cross: Yes, this is because, starting with this PR and until the end of the migration to the v2 of the API, the `instance` namespace will be in a particular state where the same commands will be generated for v1 and v2. I can't use `GetGeneratedCommands()` in both `instance/v1/custom.go` and `instance/v2alpha1/custom.go` because then the CLI will have 2 versions of `scw instance server create` for example.

For now, I didn't want to modify the existing code (already implemented commands call v1, and should keep calling v1), I just want to add support for the new template resource, which is only available in v2, so I manually picked the template-related commands using `NewCommands()` instead.
As the migration progresses, more resources will be added in v2 and removed from v1.

pypaut (approved):

remyleone (approved):

PR #6045, opened 2026-08-27
ci: add support for a prepare step before using the cache

remyleone:
Inspiration: https://github.com/rclone/rclone/commit/8b981d8a695791b16dd3acfc530c064ab0cc93b8


elwinar:
Obsolete since we reworked the CI.

PR #6117, opened 2026-09-03
feat(instance): migrate placement-group resource to API v2

Mia-Cross:


internal/namespaces/instance/v1/custom.go
```diff
@@ -196,18 +269,6 @@ func GetCommands() *core.Commands {
...
-	// Placement Group
-	//
-	human.RegisterMarshalerFunc(
-		instance.CreatePlacementGroupResponse{},
-		marshallNestedField("PlacementGroup"),
-	)
-
-	cmds.MustFind("instance", "placement-group", "create").Override(placementGroupCreateBuilder)
-	cmds.MustFind("instance", "placement-group", "get").Override(placementGroupGetBuilder)
-	cmds.MustFind("instance", "placement-group", "list").Override(placementGroupListBuilder)
```
pypaut, line 209: Just as a confirmation, the following (v1 exclusive) fonction:

```go
cmds.MustFind("instance", "placement-group", "get-servers").Override(...)
```

is not present here because the default generated function is enough?

pypaut (changes requested):

internal/namespaces/instance/v1/custom.go
Mia-Cross: Yes, this method and the other ones that are not present in v2 (`instancePlacementGroupSet()`, `instancePlacementGroupGetServers()`, `instancePlacementGroupSetServers()` and `instancePlacementGroupUpdateServers()`), never had any custom logic.

We're not sure if these methods are used or not, so we decided to keep the generated functions and deprecate them instead of just deleting them.

internal/namespaces/instance/v2alpha1/testdata/test-placement-group-get-without-servers.golden
```diff
@@ -0,0 +1,21 @@
...
+{
+  "id": "139eda29-c24d-4e6a-ab7a-24f3a60b55a9",
+  "project_id": "fa1e3217-dc80-42ac-85c3-3f034b78b552",
+  "name": "cli-pg-eloquent-banzai",
+  "policy_type": "max_availability",
+  "tags": [],
+  "created_at": "1970-01-01T00:00:00.0Z",
+  "updated_at": "1970-01-01T00:00:00.0Z",
+  "zone": "fr-par-1",
+  "servers": null
```
estellesoulard: should we just not return servers in this case? (setting an `omitempty` json tag?)

internal/namespaces/instance/v1/custom_placement_group.go
```diff
@@ -1,115 +0,0 @@
...
-
-func placementGroupListBuilder(c *core.Command) *core.Command {
-	type customListPlacementGroupsRequest struct {
-		*instance.ListPlacementGroupsRequest
-		OrganizationID *string
-		ProjectID      *string
-	}
-
-	renameOrganizationIDArgSpec(c.ArgSpecs)
-	renameProjectIDArgSpec(c.ArgSpecs)
```
estellesoulard, line 94: Kinda nit: I see in this golden cmd/scw/testdata/test-all-usage-instance-placement-group-usage.golden that we now have an `organization` and a `project` attribute. Usual namings seem to be `project-id`, `organization-id`. I'm not sure but maybe the disappearance of these lines has something to do with it? Shouldn't these renames be used in the v2 as well?

internal/namespaces/instance/v1/custom.go
pypaut: Ok 👍🏻 Shouldn't the commands be deprecated in `internal/namespaces/instance/v1/instance_cli.go` then?

Mia-Cross: Yes, although the file is generated, we could mark the commands as deprecated at generation level, but since we need to add some custom logic anyway (for the deprecation message), we might as well deprecate the command from here.

internal/namespaces/instance/v1/custom_placement_group.go
Mia-Cross: Instance v1 API, as it is one of the oldest APIs, had `organization` and `project` fields. This custom logic was needed then to be able to use `organization-id` and `project-id` on all CLI namespaces, but v2 now has the right fields so we can drop it.

The usage goldens were broken at the time of your review because the namespace is currently a mix of v1 and v2, which confuses the doc generation. I believe the issue is fixed now :+1:

PR #6132, opened 2026-09-04
feat(instance): migrate placement-group resource to API v2

Mia-Cross:
This PR is a duplicate from #6117, which was from a fork. 
Review comments are on the original PR.


internal/namespaces/instance/v1/custom_placement_group.go
```diff
@@ -1,115 +1,26 @@
...
-			request.Organization = args.OrganizationID
-			request.Project = args.ProjectID
+func placementGroupSetBuilder(c *core.Command) *core.Command {
+	return deprecateV1OnlyCommand(c)
+}
 
-			return runner(ctx, request)
-		},
-	)
+func deprecateV1OnlyCommand(c *core.Command) *core.Command {
```
pypaut, line 21: Does this function add deprecation, while also keeping `c` associated function (i.e. the auto-generated one)?

pypaut (changes requested):

internal/namespaces/instance/v1/custom_placement_group.go
Mia-Cross: That's right :+1:

pypaut: 🤌

pypaut (dismissed):

pypaut (dismissed):

cmd/scw/testdata/test-all-usage-instance-placement-group-delete-usage.golden
```diff
@@ -1,17 +1,10 @@
 🎲🎲🎲 EXIT CODE: 0 🎲🎲🎲
 🟩🟩🟩 STDOUT️ 🟩🟩🟩️
-Delete the specified placement group.
+Delete a specified placement group.
 
 USAGE:
   scw instance placement-group delete <placement-group-id ...> [arg=value ...]
 
-EXAMPLES:
```
remyleone, line 8: Could the examples be kept? They are really helpful for both agents and human

internal/namespaces/instance/v1/instance_cli_test.go
```diff
@@ -92,10 +96,12 @@ func Test_ServerUpdate(t *testing.T) {
 		AfterFunc: deleteServer("Server"),
 	}))
 
+	cmds.Merge(instanceV2.GetCommands())
```
remyleone, line 99: I'm not sure to understand why the set of commands is changed in the middle of the test.

Mia-Cross: That is because the first 2 tests only test resources from v1, the testing with actual placement-groups (now a v2 resource) starts with the third test.

cmd/scw/testdata/test-all-usage-instance-placement-group-delete-usage.golden
Mia-Cross: Examples are now migrated and updated at generation level :+1:

PR #6158, opened 2026-09-08
test(instance): cover security-group create output format

ousamabenyounes:
## What

Adds integration coverage for the current scw instance security-group create -o json response contract:

- keeps the top-level security_group wrapper;
- covers successful creation and API errors with recorded cassettes and goldens;
- uses the wrapped response in test cleanup.

## Why

The proposed unwrapping was removed after review because changing the JSON shape would break compatibility. This PR now only locks the existing behavior with regression coverage.

## Validation

- go test ./internal/namespaces/instance/v1 -run Test_SecurityGroupCreate -count=1 passes.
- Full baseline and final mise run ci replays are iso-baseline: 2,623 final tests, 17 skipped, and the same six pre-existing internal/namespaces/container/v1 Test_Deploy cassette misses. instance/v1 passes and WASM is 9/9.
- Independent code review and Claude/OCR review pass.


elwinar (changes requested):
At the moment, I'm not sure this is something we want to change. This is not strictly a bug as we don't have a convention in place on this (a quick look at the various golden files makes it obvious), so we're probably not going to change the format and break compatibility. The additional test is welcomed tho, if you want to re-submit without the fix that is something I'll be willing to approve.

ousamabenyounes:
Updated in d77ccc1f98dffdd35029c01ba324937489a1405b: removed the JSON unwrapping entirely and kept only regression coverage for the existing wrapped `security_group` response. The targeted Go test passes; GitHub’s CI workflow is awaiting maintainer approval.

ousamabenyounes:
Follow-up pushed in c348a7a6f082cdd6396e148da1f33d5c39ca32f5: deduplicated the shared security-group create command after independent review. The targeted test passes; full local CI remains iso-baseline, and Claude/OCR review passes. GitHub CI requires maintainer approval to run.

